PR #6, opened 2017-11-13
Use FCITX_INSTALL_LOCALEDIR

oahong:


wengxt:
I prefer to install file to their own prefix. unless it's necessary.

oahong:
according to cmake documentation and GNUInstallDirs.cmake, LOCALEDIR is relative to DATAROOTDIR. It should observe the prefix setting.

PR #12, opened 2017-12-03
make libuuid public for include path

HenryHu:
inputcontext_p.h includes uuid.h. As a result, anything using this header file needs the include path of libuuid. Some modules fall in this class.


wengxt:
_p.h is private.

PR #90, opened 2020-08-03
Port fcitx-diagnose

yuyichao:
This version supports both fcitx 4 and 5 and will default to warn/suggest solutions about fcitx 5 if neither is found.

Input method list is not supported yet on fcitx 5...


wengxt:
No need to support fcitx 4. Please remove it. 
fcitx4 and 5 have conflict files.

wengxt:
Can you fix Messages.sh to handle the extract I18N string for this file?

wengxt (approved):

wengxt:
surprisingly found i18n extract is already there..

PR #93, opened 2020-08-07
Implement client-side UI update dbus signal

nntae:
in preparation for fbterm support

### DIFF
--- a/src/frontend/dbusfrontend/dbusfrontend.cpp
+++ b/src/frontend/dbusfrontend/dbusfrontend.cpp
@@ -105,6 +105,48 @@ public:
         deleteSurroundingTextDBusTo(name_, offset, size);
     }
 
+    void updateClientSideUIImpl() override {
+        auto preedit =
+            im_->instance()->outputFilter(this, inputPanel().clientPreedit());
+        auto auxUp = im_->instance()->outputFilter(this, inputPanel().auxUp());
+        auto auxDown =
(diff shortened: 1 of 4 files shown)


wengxt:
So at least, can you try to make it merge-able (from web ui side)?.. I don't want to review my changes and your changes together.

Multiple ways to do it:
1. cherry-pick you commit one by one on to new master
2. reset and squash all your change into one commit.
3. rebase if it's easy to do so.

nntae:
updated to resolve conflicts

src/lib/fcitx/inputcontext.cpp
```diff
@@ -256,7 +256,11 @@ void InputContext::updatePreedit() {
 void InputContext::updateUserInterface(UserInterfaceComponent component,
                                        bool immediate) {
     FCITX_D();
-    d->emplaceEvent<InputContextUpdateUIEvent>(component, this, immediate);
```
wengxt, line 259: I think we should move this part of the code to UserInterfaceManager::flush. Thus the event is always generated.

src/frontend/dbusfrontend/dbusfrontend.cpp
```diff
@@ -23,6 +23,17 @@
 
 namespace fcitx {
 
+namespace {
+
+void buildFormattedTextVector(
```
wengxt, line 28: two choice about the naming:
1. just construct and return the vector
2. rename from build to append (if same vector is used by this function twice).

PR #129, opened 2020-10-12
Add a module that switches to the Nth input method via Hotkey

15cm:
The module allows users to config hotkeys to switch to the 1-5th input method globally/locally.

Functionality is manually verified on my local machine.

### DIFF
--- a/src/modules/imselector/imselector.cpp
+++ b/src/modules/imselector/imselector.cpp
@@ -12,6 +12,27 @@
 
 namespace fcitx {
 
+bool selectInputMethod(InputContext *ic, IMSelector *imSelector,
+                       const std::string &uniqueName, bool local) {
+    auto *state = ic->propertyFor(&imSelector->factory());
+    imSelector->instance()->setCurrentInputMethod(ic, uniqueName, local);
+    state->reset(ic);
+    imSelector->instance()->showInputMethodInformation(ic);
+    return true;
+}
+
+bool selectInputMethod(InputContext *ic, IMSelector *imSelector, size_t index,
+                       bool local) {
+    auto &inputMethodManager = imSelector->instance()->inputMethodManager();
+    auto &list = inputMethodManager.currentGroup().inputMethodList();
+    if (index >= list.size()) {
+        return false;
+    }
+    return selectInputMethod(
+        ic, imSelector,
+        inputMethodManager.entry(list[index].name())->uniqueName(), local);
+}
+
 class IMSelectorCandidateWord : public CandidateWord {
 public:
     IMSelectorCandidateWord(IMSelector *q, const InputMethodEntry *entry,
@@ -20,10 +41,7 @@ public:
           uniqueName_(entry->uniqueName()), local_(local) {}
 
     void select(InputContext *ic) const override {
-        auto *state = ic->propertyFor(&q_->factory());
-        q_->instance()->setCurrentInputMethod(ic, uniqueName_, local_);
-        state->reset(ic);
-        q_->instance()->showInputMethodInformation(ic);
+        selectInputMethod(ic, q_, uniqueName_, local_);
     }
 
     IMSelector *q_;
@@ -48,6 +66,28 @@ IMSelector::IMSelector(Instance *instance)
             }
             if (keyEvent.key().checkKeyList(config_.triggerKeyLocal.value()) &&
                 trigger(keyEvent.inputContext(), true)) {
+                keyEvent.filterAndAccept();
+                return;
+            }
+        }));
+
+    // Select input method via hotkey.
+    eventHandlers_.emplace_back(instance_->watchEvent(
+        EventType::InputContextKeyEvent, EventWatcherPhase::PreInputMethod,
+        [this](Event &event) {
+            auto &keyEvent = static_cast<KeyEvent &>(event);
+            auto *inputContext = keyEvent.inputContext();
+            if (int index =
+                    keyEvent.key().keyListIndex(config_.switchKey.value());
+                index >= 0 &&
+                selectInputMethod(inputContext, this, index, /*local=*/false)) {
+                keyEvent.filterAndAccept();
+                return;
+            }
+            if (int index =
+                    keyEvent.key().keyListIndex(config_.switchKeyLocal.value());
+                index >= 0 &&
+                selectInputMethod(inputContext, this, index, /*local=*/true)) {
                 keyEvent.filterAndAccept();
                 return;
             }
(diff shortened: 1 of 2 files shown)


wengxt:
Can you merge it with imselector?

15cm:
Sure if you think that's better. I have considered this but don't want to
blow up the imselector panel.

wengxt:
Yeah, please do so. Also you can create a section for it so it will looks more organized.

Another option is to use a single option member: KeyListOption instead of adding 5 different option.

15cm:
Done. Used the key list to bind each key list item to an input method. Added tooltips for the key list option.

Demo of the pref pane: https://i.imgur.com/YQUCofq.png

src/modules/imselector/imselector.h
```diff
@@ -25,7 +28,26 @@ FCITX_CONFIGURATION(
         "TriggerKeyLocal",
         _("Trigger Key for only current input context"),
         {},
-        KeyListConstrain()};);
+        KeyListConstrain()};
+    KeyListOptionWithToolTip switchKey{
+        this,
+        "SwitchKey",
+        _("Hotkey for switching to the Nth(0-based) input method"),
```
wengxt, line 35: N-th. also remove "(0-based)" .

src/modules/imselector/imselector.cpp
```diff
@@ -53,6 +53,27 @@ IMSelector::IMSelector(Instance *instance)
...
 
+    // Select input method via hotkey.
+    eventHandlers_.emplace_back(instance_->watchEvent(
+        EventType::InputContextKeyEvent, EventWatcherPhase::PreInputMethod,
+        [this](Event &event) {
+            auto &keyEvent = static_cast<KeyEvent &>(event);
+            if (int index =
+                    keyEvent.key().keyListIndex(config_.switchKey.value());
+                index >= 0 &&
+                selectInputMethod(keyEvent.inputContext(), index, false)) {
```
wengxt, line 64: ```/*local=*/false```

src/modules/imselector/imselector.cpp
```diff
@@ -53,6 +53,27 @@ IMSelector::IMSelector(Instance *instance)
...
+                keyEvent.filterAndAccept();
+                return;
+            }
+            if (int index =
+                    keyEvent.key().keyListIndex(config_.switchKeyLocal.value());
+                index >= 0 &&
+                selectInputMethod(keyEvent.inputContext(), index, true)) {
```
wengxt, line 71: `/*local=*/true`

src/modules/imselector/imselector.cpp
```diff
@@ -223,6 +244,21 @@ bool IMSelector::trigger(InputContext *inputContext, bool local) {
...
+bool IMSelector::selectInputMethod(InputContext *inputContext, size_t index,
+                                   bool local) {
+    auto &list =
+        instance_->inputMethodManager().currentGroup().inputMethodList();
+    if (index >= list.size())
+        return false;
+
+    // Use the same logic as selecting a input method from candidate lists. This
+    // will reset IMSelector when we select an input method via hotkey.
+    IMSelectorCandidateWord candidate_word(
```
wengxt, line 256: use camelCase variable name

src/modules/imselector/imselector.cpp
```diff
@@ -223,6 +244,21 @@ bool IMSelector::trigger(InputContext *inputContext, bool local) {
...
+        this, instance_->inputMethodManager().entry(list[index].name()), local);
+    candidate_word.select(inputContext);
```
wengxt, line 258: This implementation looks weird. You should make candidate word use a common function, instead of creating a virtual candidate word.

src/modules/imselector/imselector.h
```diff
@@ -25,7 +28,26 @@ FCITX_CONFIGURATION(
...
+        {},
+        KeyListConstrain(),
+        {},
+        ToolTipAnnotation(
+            _("Press the '+' button to bind a hotkey for next input method"))};
```
wengxt, line 40: You may explain the n-th key in this option will switch to the n-th input method here.

wengxt (changes requested):

src/modules/imselector/imselector.cpp
15cm: I choose this implementation because we already inject `IMSelector` into `IMSelectorCandidateWord`. Extracting that part of code to common utils doesn't fit well.

WDYT of removing `IMSelectorCandidateWord.IMSelector` and add a method `IMSelector::select(InputContext ic, IMSelectorCandidateWord word)`, which will call another new method `IMSelector::select(InputContext ic, string uniqueName, bool local)`. In this way we can replace this block with a shared method in `IMSelector` instead of having the selection logic in `IMSelectorCandidateWord`

src/modules/imselector/imselector.cpp
15cm: Done.

src/modules/imselector/imselector.cpp
15cm: Just realized `select` is an interface of the base class of `IMSelectorCandidateWord`. Chose the common function approach you mentioned.

src/modules/imselector/imselector.h
15cm: Done.

src/modules/imselector/imselector.cpp
15cm: Done.

src/modules/imselector/imselector.cpp
15cm: Done.

src/modules/imselector/imselector.h
15cm: Done.

15cm:
Just recalled that I should create a new commit instead of force-push for you to see the diff after conversation resolved. Will do next time.

src/modules/imselector/imselector.cpp
```diff
@@ -12,6 +12,27 @@
 
 namespace fcitx {
 
+bool selectInputMethod(InputContext *ic, IMSelector *imSelector,
```
wengxt, line 15: Actually I mean, only fcitx::Instance is required here. no need to wrap it with IMSelector

src/modules/imselector/imselector.cpp
```diff
@@ -12,6 +12,27 @@
 
 namespace fcitx {
```
wengxt, line 14: namespace {

wengxt (changes requested):

src/modules/imselector/imselector.cpp
15cm: How can we access `auto *state = ic->propertyFor(&imSelector->factory())` then?

PR #140, opened 2020-11-09
 fix showInputMethodInformation 

chenshijie-uos:
fix when only checked showInputMethodInformationWhenFocusIn it not  showInputMethodInformation


wengxt:
You can not just change the API name because it breaks ABI and API, even if now it has a slightly different meaning.

Next thing is when you make change next time, be sure you use .clang-format to format your code.

chenshijie-uos:
i will clang-format the code ,
it not only change api name,
it fix the function and it work,
may be the upstream code has been fixed,
i will use debian try again

PR #143, opened 2020-11-11
when only checked showInputMethodInformationWhenFocusIn can show information

chenshijie-uos:
fixed when only checked showInputMethodInformationWhenFocusIn can show information.

 i am not sure it is a bug,but it really make me confused.

图片


wengxt:
This is an intended change to fix #137

PR #150, opened 2020-11-20
update  zh_TW translation

chenshijie-uos:
update  zh_TW translation


wengxt:
Please use https://www.transifex.com/fcitx/

PR #151, opened 2020-11-20
fix: initialize value

buuuuuuuuuu:
initialize value


wengxt:
EventLoopException has only one constructor, and all code path will initialize error_.

PR #156, opened 2020-11-25
Fix fcitx5-diagnose.sh default env var.

p1gd0g:
Replace "fcitx" with "fcitx5", which was described in wiki.


wengxt:
No, this is kept as fcitx on purpose for maximum compatibility.
XMODFIERS doesn't matter, because it can be any value.
For im module, fcitx 5 im module can be used with both fcitx or fcitx 5.

p1gd0g:
OK~~~
The diagnose info confused me.

PR #160, opened 2020-11-28
add fcitx5-autostart.sh 

chenshijie-uos:
add fcitx5-autostart.sh


wengxt:
fcitx-autostart only exists for some historical reason.
1. dbus doesn't usually work at that time so fcitx is using private dbus. It's harder to detect if another fcitx is running.
2. sleep just another way to workaround certain bug.

Now, with systemd by default, we consider all system have dbus properly configured. So the singleton detection is solely done by dbus. No need to do this manual detection anymore, which makes fcitx5-autostart script pointless.

PR #161, opened 2020-11-28
make fcitx5-diagnose.sh able to handle the situation when env is set to `fcitx5`

karuboniru:
Don't raise confusing error message like 
```
    **Environment variable GTK_IM_MODULE is "fcitx5" instead of "fcitx". Please check if you have exported it incorrectly in any of your init files.**
    **You may have trouble using fcitx in gtk programs.**

    **Please set environment variable GTK_IM_MODULE to "fcitx" using the tool your distribution provides or add `export GTK_IM_MODULE=fcitx` to your `~/.xprofile`. See [Input Method Related Environment Variables: GTK_IM_MODULE](http://fcitx-im.org/wiki/Input_method_related_environment_variables#GTK_IM_MODULE).**
```
When those environment variable is set to `fcitx5`

### DIFF
--- a/data/fcitx5-diagnose.sh
+++ b/data/fcitx5-diagnose.sh
@@ -1087,6 +1087,10 @@ check_xim() {
     elif [ "${XMODIFIERS}" = '@im=fcitx' ]; then
         _env_correct 'XMODIFIERS' '@im=fcitx'
         __need_blank_line=0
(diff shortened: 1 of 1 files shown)


wengxt:
No. we want people to keep using fcitx.

whriedplanck:
So fast ......

PR #167, opened 2020-12-07
Fix build of fcitx5 on *BSD systems

### LINKED ISSUE #165: Build failure on FreeBSD
**Describe the bug**
Latest code does not build on FreeBSD.

```
/home/henryhu/proj/fcitx5/src/lib/fcitx-utils/misc.cpp:106:12: error: use of undeclared identifier 'result'
    return result;
           ^

```

**To Reproduce**
1. Clone latest source on a FreeBSD system
2. run cmake & build

**Expected behavior**
It should build and work normally.

**Desktop (please complete the following information):**
 - Desktop: KDE
 - Display server type: X11

**Additional context**
Seems like that the code depends on libkvm for *BSD systems; it checks for LIBKVM_FOUND. On the other hand, the configuration script never searched for libkvm.

HenryHu:
On *BSD systems, functionalities of getProcessName() is implemented using libkvm. This was several months ago.
However, libkvm was not searched for in CMakeLists, and the relevant include directories and libraries are not added during the config. Moreover, the needed headers are not included, if libkvm is needed.
This request adds 2 commits which adds the missing bits. These are mostly copied from fcitx4. I've built the latest fcitx5 with it on FreeBSD, and used it successfully. This should fix #165


cmake/FindLibkvm.cmake
```diff
@@ -0,0 +1,15 @@
...
+#
+#  LIBKVM_FOUND - system has Libkvm
+#  LIBKVM_INCLUDE_DIR - Libkvm include directory
+#  LIBKVM_LIBRARIES - Libraries needed to use Libkvm
+#
+
+find_path(LIBKVM_INCLUDE_DIR kvm.h)
+find_library(LIBKVM_LIBRARIES NAMES kvm)
+
+include(FindPackageHandleStandardArgs)
```
wengxt, line 12: Can you make this using modern cmake feature?
See https://github.com/fcitx/fcitx5/blob/5c0e52b4e8410de3c300feeff28ec78c7be056ac/cmake/FindDL.cmake#L40 for an example

CMakeLists.txt
```diff
@@ -71,6 +71,9 @@ pkg_get_variable(DBUS_SYSTEM_BUS_DEFAULT_ADDRESS "dbus-1" "system_bus_default_ad
 pkg_check_modules(LibEvent REQUIRED IMPORTED_TARGET "libevent")
 endif()
 
+if(${CMAKE_SYSTEM_NAME} MATCHES "BSD")
+  find_package(Libkvm REQUIRED)
```
wengxt, line 75: LibKVM would be better

CMakeLists.txt
```diff
@@ -71,6 +71,9 @@ pkg_get_variable(DBUS_SYSTEM_BUS_DEFAULT_ADDRESS "dbus-1" "system_bus_default_ad
...
+endif(${CMAKE_SYSTEM_NAME} MATCHES "BSD")
```
wengxt, line 76: now endif() is preferred.

wengxt (changes requested):

CMakeLists.txt
HenryHu: Done.

CMakeLists.txt
HenryHu: Done.

cmake/FindLibkvm.cmake
HenryHu: Done.

HenryHu:
Updated according to comments. PTAL.

PR #172, opened 2020-12-10
FTBFS: gettext < 0.20

marguerite:
With gettext 0.19.8.1(the last version before 0.20), the build will fail with message:

    msgfmt: cannot locate ITS rules for org.fcitx.Fcitx5.metainfo.xml.in

So I provide such rules for gettext < 0.20. rules are copyed from github.com/hughsie/appstream-glib.

This will fix build for openSUSE Leap 15.2


wengxt:
Ah, you dont need this, add this to dependency would resolve it.

https://opensuse.pkgs.org/15.1/opensuse-oss-x86_64/appstream-glib-devel-0.7.7-lp151.2.3.x86_64.rpm.html

PR #183, opened 2020-12-21
Improve supports for Apple platform

Inokinoki:
Add `getProcessName` support for Apple platform (BSD-like, but without libkvm).

Fix `modifiedTime` and explicitly export `std::timespec` from `struct timespec` on Apple platform.

### DIFF
--- a/src/lib/fcitx/icontheme.cpp
+++ b/src/lib/fcitx/icontheme.cpp
@@ -19,6 +19,13 @@
 #include "fcitx-utils/log.h"
 #include "config.h"
 
+#ifdef __APPLE__
+// Apple SDKs don't define ::timespec_get and ::timespec unconditionally in C++ mode.
+namespace std {
+    using ::timespec;
+};
+#endif
+
 namespace fcitx {
 
 namespace {
@@ -31,11 +38,11 @@ modifiedTime(const T &p) {
     return t;
 }
 template <typename T>
-inline std::enable_if_t<(&T::st_mtimespec), std::timespec>
+inline std::enable_if_t<(&T::st_mtimespec, true), std::timespec>
 modifiedTime(const T &p) {
     std::timespec t;
     t.tv_sec = p.st_mtimespec.tv_sec;
-    t.tv_nsec = p.st_mtimensec.tv_nsec;
+    t.tv_nsec = p.st_mtimespec.tv_nsec;
     return t;
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/fcitx/icontheme.cpp
```diff
@@ -19,6 +19,13 @@
 #include "fcitx-utils/log.h"
 #include "config.h"
 
+#ifdef __APPLE__
```
wengxt, line 22: Are you sure? it is said that says that std::timespec should be exported by C++17 via `<ctime>`, I think mac should use  llvm libc++ anyway?

wengxt (changes requested):

src/lib/fcitx/icontheme.cpp
Inokinoki: Yes.

Yet in the current version of libcxx (LLVM libc++), the definition is here:

https://github.com/llvm/llvm-project/blob/38ca7face67e8488d482b66a999d0a685806879f/libcxx/include/ctime#L75

and [the comments](https://github.com/llvm/llvm-project/blob/38ca7face67e8488d482b66a999d0a685806879f/libcxx/include/ctime#L56
) in the same file:

```
// Apple SDKs don't define ::timespec_get unconditionally in C++ mode. This
// should be fixed in future SDKs, but for the time being we need to avoid
// trying to use that declaration when the SDK doesn't provide it.
```

telling that the current SDKs of Apple do not export them.


The build error is as follow:

```
/Users/***/CraftRoot/download/git/libs/fcitx5/src/lib/fcitx/icontheme.cpp:27:51: error: no member named 'timespec' in namespace 'std'
inline std::enable_if_t<(&T::st_mtim, true), std::timespec>
                                             ~~~~~^
/Users/***/CraftRoot/download/git/libs/fcitx5/src/lib/fcitx/icontheme.cpp:29:5: error: no type named 'timespec' in namespace 'std'; did you mean simply 'timespec'?
    std::timespec t;
    ^~~~~~~~~~~~~
    timespec
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/sys/_types/_timespec.h:33:1: note: 'timespec' declared here
_STRUCT_TIMESPEC
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/sys/_types/_timespec.h:29:40: note: expanded from macro '_STRUCT_TIMESPEC'
#define _STRUCT_TIMESPEC        struct timespec
```

PR #201, opened 2021-01-10
Add fcitx4frontend

vifly:


src/frontend/fcitx4frontend/fcitx4frontend.h
```diff
@@ -0,0 +1,45 @@
...
+    dbus::Bus *bus();
+    Instance *instance() { return instance_; }
+    int nextIcIdx() { return ++icIdx_; }
+
+private:
+    FCITX_ADDON_DEPENDENCY_LOADER(dbus, instance_->addonManager());
+
+    Instance *instance_;
+    std::unique_ptr<dbus::Bus> portalBus_;
+    std::unique_ptr<Fcitx4InputMethod> Fcitx4InputMethod_;
```
wengxt, line 38: lowerCase

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,357 @@
...
+    if (content.empty()) {
+        content = "machine-id";
+    }
+
+    return content;
+}
+
+Fcitx4FrontendModule::Fcitx4FrontendModule(Instance *instance)
+    : instance_(instance),
+      portalBus_(std::make_unique<dbus::Bus>(dbus::BusType::Session)),
```
wengxt, line 303: is this ever used?

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,357 @@
...
+    Flags<dbus::RequestNameFlag> requestFlag =
+        dbus::RequestNameFlag::ReplaceExisting;
+    this->bus()->requestName(FCITX_DBUS_SERVICE, requestFlag);
+    auto display = getDisplayNumber();
+    auto dbusServiceName =
+        stringutils::concat(FCITX_DBUS_SERVICE, "-", display);
+    this->bus()->requestName(dbusServiceName, requestFlag);
+
+    auto localMachineId = getLocalMachineId();
+    auto path = stringutils::joinPath(
```
wengxt, line 317: use StandardPath::safeSave and check for failure.

You may simply throw exception on failure.

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,357 @@
...
+    fwrite(&curPid, sizeof(pid_t), 1, fp);
+    fwrite(&curPid, sizeof(pid_t), 1, fp);
+    fclose(fp);
+
+    event_ = instance_->watchEvent(
+        EventType::InputContextInputMethodActivated, EventWatcherPhase::Default,
+        [this](Event &event) {
+            auto &activated = static_cast<InputMethodActivatedEvent &>(event);
+            auto *ic = activated.inputContext();
+            if (strcmp(ic->frontend(), "fcitx4frontend") == 0) {
```
wengxt, line 335: I suggest use "fcitx4" as name string instead.

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,357 @@
...
+              [this](const std::string &, const std::string &,
+                     const std::string &newName) {
+                  if (newName.empty()) {
+                      delete this;
+                  }
+              })),
+          name_(sender) {
+        processKeyEventMethod.setClosureFunction(
+            [this](dbus::Message message, const dbus::ObjectMethod &method) {
+                if (capabilityFlags().test(CapabilityFlag::KeyEventOrderFix)) {
```
wengxt, line 86: fcitx4 doesn't have this bit flag, so you can assume it is always false

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,357 @@
...
+    auto dbusServiceName =
+        stringutils::concat(FCITX_DBUS_SERVICE, "-", display);
+    this->bus()->requestName(dbusServiceName, requestFlag);
+
+    auto localMachineId = getLocalMachineId();
+    auto path = stringutils::joinPath(
+        StandardPath().userDirectory(StandardPath::Type::Config), "fcitx",
+        "dbus", stringutils::concat(localMachineId, "-", display));
+    auto *fp = fopen(path.c_str(), "w");
+    fprintf(fp, "%s", this->bus()->address().c_str());
```
wengxt, line 321: use fwrite here, c_str() is always return a pointer to nul terminated string.

wengxt (changes requested):

src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+namespace fcitx {
+
+namespace {
+
+std::vector<dbus::DBusStruct<std::string, int>>
+buildFormattedTextVector(const Text &text) {
+    std::vector<dbus::DBusStruct<std::string, int>> vector;
+    for (int i = 0, e = text.size(); i < e; i++) {
+        vector.emplace_back(std::make_tuple(
```
wengxt, line 35: While fcitx4's format flag is almost compatible:

There's one difference: (1<<3) bit means "no underline", while in fcitx 5 it means "underline". You need to flip this bit.

wengxt (changes requested):

src/frontend/fcitx4frontend/fcitx4frontend.cpp
wengxt: See https://github.com/fcitx/fcitx/blob/38107d8454ec503dc9d93cf5f8e6e5ee7be88857/src/frontend/ipcportal/ipcportal.c#L714

Issue #208, opened 2021-01-19
Failed to build on alpha and GNU hurd


zhsj:
Alpha https://buildd.debian.org/status/fetch.php?pkg=fcitx5&arch=alpha&ver=5.0.4-2&stamp=1611053086&raw=0

```
/<<PKGBUILDDIR>>/src/lib/fcitx/icontheme.cpp:227:41: error: call of overloaded ‘modifiedTime(stat&)’ is ambiguous
  227 |         if (timespecLess(modifiedTime(st), modifiedTime(dirSt))) {
      |                                         ^

```
It seems to work at 5.0.3.

Hurd https://buildd.debian.org/status/fetch.php?pkg=fcitx5&arch=hurd-i386&ver=5.0.4-2&stamp=1611032751&raw=0

```
/<<PKGBUILDDIR>>/src/lib/fcitx-utils/misc.cpp:125:12: error: ‘result’ was not declared in this scope
  125 |     return result;
      |            ^~~~~~
```

The error seems not hard to resolve and I may have time on weekend to investigate.

PR #210, opened 2021-01-21
Fix build on GNU Hurd

zhsj:
On https://www.gnu.org/software/hurd/hurd/porting/guidelines.html
`__GNU__` can be used for GNU Hurd.

https://www.gnu.org/software/hurd/hurd/translator/procfs.html
And it has /proc/<pid/exe.

For #208


zhsj:
It just builds, but the tests doesn't pass...

wengxt:
I think we can always fallback to procfs, without checking the all the possible special case

zhsj:
> I think we can always fallback to procfs, without checking the all the possible special case

Done.

PR #234, opened 2021-02-27
Compatible with low versions gcc

chenshijie-uos:


wengxt:
Please use clang instead on such platform.

PR #274, opened 2021-05-07
Do not start key repeat timer if event is redirected back to wayland client

yuja:
While debugging sway/wlroot issue, I noticed key repeat events are duplicated.
The event came from fcitx5 restarts the client-side repeat timer, which
actually slows down the key-repeat interval. Since key repeating should be
handled by client on Wayland, IM shouldn't emulate it if the event isn't
consumed by IM and sent back to the client through virtual keyboard.

I've tested the V2 server on Sway. I also found a similar code in V1, so
applied a similar fix assuming that the returned keysym would be processed
as a key event by client.

Because I'm not sure if the event variable can be modified by keyEvent(event),
I've added timeEvent_->setEnabled(false). If the event variable can be
considered a constant, it's probably better to postpone the timer configuration
at all.

### DIFF
--- a/src/frontend/waylandim/waylandimserverv2.cpp
+++ b/src/frontend/waylandim/waylandimserverv2.cpp
@@ -198,8 +198,7 @@ void WaylandIMInputContextV2::repeat() {
         false, repeatTime_);
     if (!keyEvent(event)) {
         vk_->key(repeatTime_, event.rawKey().code() - 8,
-                 event.isRelease() ? WL_KEYBOARD_KEY_STATE_RELEASED
-                                   : WL_KEYBOARD_KEY_STATE_PRESSED);
+                 WL_KEYBOARD_KEY_STATE_PRESSED);
     }
 
     timeEvent_->setNextInterval(1000000 / repeatRate_);
--- a/src/frontend/waylandim/waylandimserverv2.cpp
+++ b/src/frontend/waylandim/waylandimserverv2.cpp
@@ -197,8 +197,16 @@ void WaylandIMInputContextV2::repeat() {
         Key(repeatSym_, server_->modifiers_ | KeyState::Repeat, repeatKey_ + 8),
         false, repeatTime_);
     if (!keyEvent(event)) {
-        vk_->key(repeatTime_, event.rawKey().code() - 8,
-                 WL_KEYBOARD_KEY_STATE_PRESSED);
+        // Since key repeating is also handled by Wayland client, key pressed
+        // event shouldn't be generated repeatedly. Otherwise, the client-side
+        // repetition would be interleaved with the IM event.
+        if (!lastPressedSentToVk_) {
+            vk_->key(repeatTime_, event.rawKey().code() - 8,
+                     WL_KEYBOARD_KEY_STATE_PRESSED);
+            lastPressedSentToVk_ = true;
+        }
+    } else {
+        lastPressedSentToVk_ = false;
     }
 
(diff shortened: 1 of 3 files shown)


wengxt:
While this makes some sense, it actually makes an assumption:
All the future "repeated" key will not be handled by IM. In reality, this is not the case and we have a feature make use of this.

And in V1, the keysym is not going through "virtual keyboard", so if any client key-repetition is triggered by that, it's not correct. but after all there's no real v1 client so we don't actually know which is correct.

yuja:
> All the future "repeated" key will not be handled by IM. In reality, this is not the case and we have a feature make use of this.

Good point. So we can suppress the IM-side repeats only if event[n - 1] was vk press and event[n] is also vk press.
I'll update the patch shortly.

> And in V1, the keysym is not going through "virtual keyboard", so if any client key-repetition is triggered by that, it's not correct.

Thanks for the clarification. It was unclear for me how client should process keysym event.

src/frontend/waylandim/waylandimserverv2.cpp
```diff
@@ -197,9 +197,17 @@ void WaylandIMInputContextV2::repeat() {
...
-                 event.isRelease() ? WL_KEYBOARD_KEY_STATE_RELEASED
-                                   : WL_KEYBOARD_KEY_STATE_PRESSED);
+        // Since key repeating is also handled by Wayland client, key pressed
+        // event shouldn't be generated repeatedly. Otherwise, the client-side
+        // repetition would be interleaved with the IM event.
+        if (!lastPressedSentToVk_ || event.isRelease()) {
+            vk_->key(repeatTime_, event.rawKey().code() - 8,
+                     event.isRelease() ? WL_KEYBOARD_KEY_STATE_RELEASED
+                                       : WL_KEYBOARD_KEY_STATE_PRESSED);
+            lastPressedSentToVk_ = !event.isRelease();
```
wengxt, line 207: event.isRelease() is always false.

wengxt:
Actually I mean the problem is unsolvable due to the protocol design flaw.

Base on this I filed a issue: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/46

I'd say in this patch we just try to do the thing cause least problem.

if a key is forward to vk, just stop key repetition on input method side, though the remaining key repetition created by the client won't be forwarded to input method, which is not we want to see. The logic should be really simple then.

wengxt:
Actually I tried to test it on sway. Though there's some issue with alacritty, gtk4-demo-application appear to be quite normal. I don't think this is actually needed. What's the application that has problem?

yuja:
> Though there's some issue with alacritty, gtk4-demo-application appear to be quite normal.

Yep, wayland-native terminal emulator had some issue. In my case, it was foot. AFAIK, Gnome
and KDE/Qt apps have their own IM bridges so these apps should work without problem.

Anyway, the protocol has design flow as you mentioned, and I agree to drop this patch.

wengxt:
@yuja Please try the latest commit, I think I found a way to hack through.

wengxt:
So fcitx is only relevant to text typing, so by default it should assume key repetition should happen already.

Also, in X11 world there's two different mode of repetition: XkbSetDetectableAutoRepeat
https://linux.die.net/man/3/xkbsetdetectableautorepeat
- press release interleaving
-
Continuous press and only one release. Gtk and Qt actually uses different mode by default under X11.

Also, fcitx would prevent client side repetition already if the inital press event is filtered by fcitx. The application won't know the key is pressed in the first place. So to sum it up, it's likely to be Ok.

PR #301, opened 2021-06-27
add support for Android Logcat

vifly:


wengxt:
I suggest keep this code in the fcitx5-android project. I already add Log::setLogStream specifically for this.

PR #308, opened 2021-07-07
Fix crash due to undefined behaviours

sgn:
I discovered this problem when running `fcitx5-config-qt`, `fcitx5` crashes as soon as the config tool is launched.

I have gcc 10.2.1 (pre-version), fcitx5 (5.0.8), dbus 1.12.20.

-----

As of it's now, we're downcasting from SignalBase to
Signal<SignalType> (which is the same type with
Signal<SignalType, LastValue<...>>) in its member functions.

The downcast is incorrect if its Combiner is NOT LastValue, fcitx5 will
run into undefined behaviours with this kind of casting.

In my local machine, this result in indeterminated value in
CheckUpdateResult, then fcitx5 will crash because of DBus' assertion.

When compiling with: `-fsanitize=undefined`, this problem is being
reported:

```
../src/lib/fcitx/../fcitx-utils/connectableobject.h:115:18: runtime error: downcast of address 0x559bd4a2a5b0 which does not point to an object of type 'Signal'
0x559bd4a2a5b0: note: object is of type '*N5fcitx6SignalIFbvENS_12_GLOBAL__N_117CheckUpdateResultEEE'
 00 00 00 00  40 25 21 2e f7 7f 00 00  e0 f5 a2 d4 9b 55 00 00  20 70 61 67 65 00 00 00  41 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for '*N5fcitx6SignalIFbvENS_12_GLOBAL__N_117CheckUpdateResultEEE'
../src/lib/fcitx/../fcitx-utils/connectableobject.h:116:21: runtime error: member call on address 0x559bd4a2a5b0 which does not point to an object of type 'Signal'
0x559bd4a2a5b0: note: object is of type '*N5fcitx6SignalIFbvENS_12_GLOBAL__N_117CheckUpdateResultEEE'
 00 00 00 00  40 25 21 2e f7 7f 00 00  e0 f5 a2 d4 9b 55 00 00  20 70 61 67 65 00 00 00  41 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for '*N5fcitx6SignalIFbvENS_12_GLOBAL__N_117CheckUpdateResultEEE'
```

### DIFF
--- a/src/lib/fcitx-utils/connectableobject.h
+++ b/src/lib/fcitx-utils/connectableobject.h
@@ -48,20 +48,6 @@
 #define FCITX_DEFINE_SIGNAL_PRIVATE(CLASS_NAME, NAME)                          \
     ::fcitx::SignalAdaptor<CLASS_NAME::NAME> CLASS_NAME##NAME##Adaptor { q_ptr }
 
-/// \brief Declare a signal in pimpl class.
-#define FCITX_DEFINE_SIGNAL_WITH_COMBINER(CLASS_NAME, NAME, COMBINER)          \
-    ::fcitx::SignalAdaptor<CLASS_NAME::NAME, COMBINER>                         \
-        CLASS_NAME##NAME##Adaptor {                                            \
-        this                                                                   \
-    }
-
-/// \brief Declare a signal in pimpl class.
-#define FCITX_DEFINE_SIGNAL_PRIVATE_WITH_COMBINER(CLASS_NAME, NAME, COMBINER)  \
-    ::fcitx::SignalAdaptor<CLASS_NAME::NAME, COMBINER>                         \
-        CLASS_NAME##NAME##Adaptor {                                            \
-        q_ptr                                                                  \
-    }
-
 namespace fcitx {
 
 class ConnectableObject;
--- a/src/lib/fcitx-utils/connectableobject.h
+++ b/src/lib/fcitx-utils/connectableobject.h
@@ -69,6 +69,8 @@ class ConnectableObject;
 /// \brief Helper class to register class.
 template <typename T, typename Combiner = typename T::combinerType>
 class SignalAdaptor {
+    // FIXME remove Combiner when we can break ABI.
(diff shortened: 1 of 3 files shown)


wengxt:
Thanks, but I'd just avoid use this for now until we bump next abi version.

sgn:
Sorry, which ABI breakage did you mean?
In my naive check:
- all data structure layout isn't changed.
- the symbol table for old and new fcitx-utils only differs in `_ZN5fcitx12_GLOBAL__N_115checkBoolEnvVarEPKc` which is function in anonymous namespace, so nothing to worry about 
[libfcitx-utils-abi.old.txt](https://github.com/fcitx/fcitx5/files/6937298/libfcitx-utils-abi.old.txt)
[libfcitx-utils-abi.new.txt](https://github.com/fcitx/fcitx5/files/6937300/libfcitx-utils-abi.new.txt)

- new `libFcitxCore.so`'s  only misses `fcitx::Signal<bool (), fcitx::LastValue<bool> >::operator()()` and those typeinfo for some `std::function`, which could be added in a dummy functions 
[libfcitx-abi.old.txt](https://github.com/fcitx/fcitx5/files/6937330/libfcitx-abi.old.txt)
[libfcitx-abi.new.txt](https://github.com/fcitx/fcitx5/files/6937335/libfcitx-abi.new.txt)

---

Anyway, this is not a deal breaker since I think current CheckUpdateResult isn't used anyway.

wengxt:
Here's the thing, with your change, you can't do
```
instance->connect<CheckUpdate>(...)
```
That defeat the purpose of this signal connect API.

The intention of this "combiner" should be able to erase and not expose the combiner type in the symbol, which means it would require change to ConnectableObject's template function compiled within each addon library. 

Ideally, there should be a type erasured version of Signal<function>, but then you introduce the change to SignalBase -> Signal type hierarchy.

wengxt:
Same feature was revived via a different mechanism in 5ddc4eae932eef21405bf1cd71ed37b8f37ed9c8

PR #313, opened 2021-07-29
Fix typo

duongdominhchau:


src/ui/classic/classicui.h
```diff
@@ -102,7 +102,7 @@ FCITX_CONFIGURATION(
         useInputMethodLanguageToDisplayText{
             this,
             "UseInputMethodLangaugeToDisplayText",
```
flyxyz123, line 104: should we also change the typo in "UseInputMethodLangaugeToDisplayText"? change to "UseInputMethodLanguageToDisplayText"?

PR #328, opened 2021-08-23
Fix a bug that wayland module cannot input text to weston-editor

ashie:
According to the protocol specification,
`zwp_input_method_context_v1::key` has to be used for unhandled key
events:

  https://github.com/wayland-project/wayland-protocols/blob/7dffa6f346ae32f7888177d74a45459997059767/unstable/input-method/input-method-unstable-v1.xml#L171-L176

But fcitx5 uses `zwp_input_method_context_v1::keysym` instead. So that it
cannot input text to weston-editor when IM is disabled.

Signed-off-by: Takuro Ashie <ashie@clear-code.com>


ashie:
BTW we are trying to use Fcitx5 with Chromium with wayland backend on Weston.
It's works very fine than other IMs except this issue, thanks for your great work!

FYI: Here is a captured movie works with Chromium on Weston:
https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/679

PR #343, opened 2021-09-27
Update localization for zh_TW.

zonble:


wengxt:
Please use transifex instead https://www.transifex.com/fcitx/fcitx/

PR #390, opened 2021-11-26
Override locale dir by environment variable.

rocka:
Able to override locale path by environment variable `FCITX_LOCALE_DIR`, in case locale files cannot be located by compile-time variable `FCITX_INSTALL_LOCALEDIR` (namely Android).

### DIFF
--- a/src/lib/fcitx-utils/i18n.cpp
+++ b/src/lib/fcitx-utils/i18n.cpp
@@ -22,8 +22,11 @@ public:
         if (domains_.count(domain)) {
             return;
         }
-        const auto *localedir = StandardPath::fcitxPath("localedir");
         if (!dir) {
+            const auto *localedir = getenv("FCITX_LOCALE_DIR");
+            if (!localedir) {
+                localedir = StandardPath::fcitxPath("localedir");
+            }
             dir = localedir;
         }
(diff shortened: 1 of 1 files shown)


src/lib/fcitx-utils/i18n.cpp
```diff
@@ -22,8 +22,11 @@ class GettextManager {
         if (domains_.count(domain)) {
             return;
         }
-        const auto *localedir = StandardPath::fcitxPath("localedir");
         if (!dir) {
+            const auto *localedir = getenv("FCITX_LOCALE_DIR");
```
wengxt, line 26: I suggest you make FCITX_LOCALE_DIR to overrides everything regardless what is being passed to registerDomain.

wengxt (changes requested):

src/lib/fcitx-utils/i18n.cpp
rocka: Would it affet fcitx5 itself? I can't find call to `registerDomain("fcitx5", ...)` in the repo ...
But if I call `registerDomain("fcitx5", ...)` manually before it's ready, translations will work as expected.

wengxt:
this is not required anymore.

PR #394, opened 2021-11-30
Export unicode addon's trigger function.

rocka:
Allow triggering unicode input mode in other addons, for more convenient usage (namely on Android).

### DIFF
--- a/src/modules/unicode/unicode.cpp
+++ b/src/modules/unicode/unicode.cpp
@@ -78,7 +78,9 @@ Unicode::Unicode(Instance *instance)
                 return;
             }
             if (keyEvent.key().checkKeyList(*config_.triggerKey)) {
-                trigger(keyEvent.inputContext());
+                if (!trigger(keyEvent.inputContext())) {
+                    return;
+                }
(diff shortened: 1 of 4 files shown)


src/modules/unicode/unicode_public.h
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-FileCopyrightText: 2021-2021 CSSlayer <wengxt@gmail.com>
```
wengxt, line 2: you should use your own copyright text.

src/modules/unicode/unicode.cpp
```diff
@@ -231,6 +230,7 @@ Unicode::Unicode(Instance *instance)
 Unicode::~Unicode() {}
 
 void Unicode::trigger(InputContext *inputContext) {
+    if (!data_.load()) return;
```
wengxt, line 233: I suggest to make trigger() has a bool return value so it can preserve the old behavior. Though it's unlikely to happen because the data file is not really intended to be override by others.

wengxt (changes requested):

PR #401, opened 2021-12-07
Add touch event to WaylandWindow

daipom:
`WaylandWindow` has `click` event, but doesn't have `touch` event.
It is possible to add the event by using `WlSeat::getTouch`.

I understand that more implementation is needed to manage touch events in earnest,
but I think these modifications are necessary to make touch event work at least first.

### DIFF
--- a/src/ui/classic/waylandpointer.cpp
+++ b/src/ui/classic/waylandpointer.cpp
@@ -23,7 +23,7 @@ WaylandPointer::WaylandPointer(wayland::WlSeat *seat) {
         if ((caps & WL_SEAT_CAPABILITY_TOUCH) && !touch_) {
             touch_.reset(seat->getTouch());
             initTouch();
-        } else if (!(caps & WL_SEAT_CAPABILITY_POINTER) && touch_) {
+        } else if (!(caps & WL_SEAT_CAPABILITY_TOUCH) && touch_) {
             touch_.reset();
         }
     });
--- a/src/ui/classic/waylandpointer.cpp
+++ b/src/ui/classic/waylandpointer.cpp
@@ -36,35 +36,35 @@ void WaylandPointer::initPointer() {
         if (!window) {
             return;
         }
-        focus_ = window->watch();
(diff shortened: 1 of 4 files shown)


wengxt (changes requested):

src/ui/classic/waylandpointer.cpp
```diff
@@ -19,6 +19,13 @@ WaylandPointer::WaylandPointer(wayland::WlSeat *seat) {
         } else if (!(caps & WL_SEAT_CAPABILITY_POINTER) && pointer_) {
             pointer_.reset();
         }
+
+        if ((caps & WL_SEAT_CAPABILITY_TOUCH) && !touch_) {
+            touch_.reset(seat->getTouch());
+            initTouch();
+        } else if (!(caps & WL_SEAT_CAPABILITY_POINTER) && touch_) {
```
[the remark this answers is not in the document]
daipom: Thank you! I have fixed this.

src/ui/classic/waylandpointer.cpp
```diff
@@ -62,4 +69,28 @@ void WaylandPointer::initPointer() {
...
 
+void WaylandPointer::initTouch() {
+    touch_->down().connect(
+        [this](uint32_t, uint32_t, wayland::WlSurface *surface, int, wl_fixed_t sx, wl_fixed_t sy) {
+            // TODO handle id for multiple touch
+            auto *window = static_cast<WaylandWindow *>(surface->userData());
+            if (!window) {
+                return;
+            }
+            focus_ = window->watch();
```
wengxt, line 80: touch focus should not be the same as pointer foucs. A new focus member that is only used by touch events should be defined.

wengxt (changes requested):

src/ui/classic/waylandpointer.cpp
daipom: Thank you!
I have made new focus members for touch events.

daipom:
Thank you for reviewing and merging!

PR #409, opened 2021-12-14
Upgrade wl_output to v4

tonyfettes:
Fcitx5 crashed due to
```
listener function for opcode 4 of wl_output is NULL
```
so I copy-pasted code and changed some variable names to make it fit to wl_output v4.

See https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/protocol/wayland.xml

### DIFF
--- a/src/lib/fcitx-wayland/core/wl_output.cpp
+++ b/src/lib/fcitx-wayland/core/wl_output.cpp
@@ -28,6 +28,16 @@ const struct wl_output_listener WlOutput::listener = {
         assert(*obj == wldata);
         { return obj->scale()(factor); }
     },
+    [](void *data, wl_output *wldata, const char *name) {
+      auto *obj = static_cast<WlOutput *>(data);
+      assert(*obj == wldata);
+      { return obj->name()(name); }
+    },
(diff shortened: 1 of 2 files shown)


wengxt:
while we could certainly do that, the thing is why would we want to do that? The output name is pretty useless to fcitx use case. Also, the down side of doing this is we would increase the bar of dependency and it won’t build anymore on the old system.

tonyfettes:
Yeah, I'm just thinking about bind it to a specific versions... Would that be acceptable?

wengxt:
I think what we might want to do here is to request a lower version instead of adding a new one, I’ll try to take a look. I don’t remember wayland has such requirement to define all the listener functions.

tonyfettes:
> I think what we might want to do here is to request a lower version instead of adding a new one

I did this to some of my other applications and it worked flawlessly. But fcitx5 has a relatively large code base so I failed to figure it out.

> I’ll try to take a look.

Thanks a lot.

wengxt:
@tonyfettes can you give the commit above a try? ^

tonyfettes:
@wengxt Yes, that works.

PR #410, opened 2021-12-14
add XDG_SESSION_TYPE

xingwozhonghua:
add XDG_SESSION_TYPE


xingwozhonghua:
https://github.com/linuxdeepin/fcitx/commit/f5791db7c6fa5415e1b7dbb6b2d6f08f58627c38

```
CPUINFO=$(grep "model name" /proc/cpuinfo | awk -F[:@] '{print $2}' | sed 's/ \+$//g'| uniq) 
if [[ $CPUINFO == "*Loongson*" ]]; then
    sleep 17
else
    sleep 9
fi
```
想添加一个cpu的判断。

PR #417, opened 2021-12-24
Fix a problem where spellhint by enchant does not work

daipom:
It's a simple bug.
Because of this bug, the prediction candidates by `enchant` did not show up.
This fix makes the candidates to show up correctly.


wengxt:
Thanks!

daipom:
Thank you for merging!

PR #420, opened 2021-12-27
Fix comparison warning

daipom:
I'm sorry.
The fix of https://github.com/fcitx/fcitx5/pull/417 causes a comparison warning.

```
warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’
```

I have fixed this.


daipom:
@wengxt Thank you for merging.

PR #428, opened 2022-01-11
`CMAKE_PREFIX_PATH` should have higer priority than hard codeded paths

ashie:
In the current inmplementation, hard codeded addon config directories
(/usr/local/share/fcitx5/addon:/usr/share/fcitx5/addon) have higher
priority than the directory specified by `CMAKE_PREFIX_PATH` on loading
addon configs. As a result fcitx5 installed in non-regular paths such
as `/opt/fcitx5` fails to load addons when the file name described in
`Library` in an addon config file is different with the one described
in /usr/share/fcitx5/addon/*.conf.
Because fcitx5 v5.0.11 adds `lib` prefix to each bundled addons, it will
occur when distrubution's fcitx5 is v5.0.10 or former.

### DIFF
--- a/src/lib/fcitx-utils/standardpath.cpp
+++ b/src/lib/fcitx-utils/standardpath.cpp
@@ -190,10 +190,11 @@ private:
 
     static std::vector<std::string> defaultPaths(const char *env,
                                                  const char *defaultPath,
-                                                 const char *fcitxPath) {
+                                                 const char *fcitxPathType) {
         std::vector<std::string> dirs;
 
-        const char *dir = getenv(env);
+        const char *envDir = getenv(env);
+        const char *dir = envDir;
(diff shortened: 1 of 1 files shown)


wengxt:
No, this is intended behavior. It is intended that fcitx path is the last fallback.

With your change it will cause problem is prefix=/usr . The look up order will then be ~/.local/share, /usr/share /usr/local/share. And there's no way to fix it to follow xdg spec.

As you can see here, the rationale here is that "fcitx prefix" is the extension of the standard XDG spec. XDG spec will always be the first one to consider, if it doesn't find any file at all, then fallback to Fcitx prefix.

You can set XDG_DATA_DIRS to before running your own fcitx5.

ashie:
I got it, thanks!

PR #429, opened 2022-01-11
Fix build with ENABLE_KEYBOARD=off.

rocka:
Changes below

https://github.com/fcitx/fcitx5/blob/09969b0982cca27a892c7bab17a1c33df435280f/src/lib/fcitx/instance.cpp#L1910-L1915

added in 217593a breaks building with `ENABLE_KEYBOARD=off` (namely on Android). Also those options seems useless without XKB.

### DIFF
--- a/src/lib/fcitx/globalconfig.cpp
+++ b/src/lib/fcitx/globalconfig.cpp
@@ -151,19 +151,30 @@ FCITX_CONFIGURATION(
                                               _("Default page size"), 5,
                                               IntConstrain(1, 10)};
 #ifdef ENABLE_KEYBOARD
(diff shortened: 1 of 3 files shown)


src/lib/fcitx/globalconfig.cpp
```diff
@@ -150,6 +150,7 @@ FCITX_CONFIGURATION(
     Option<int, IntConstrain> defaultPageSize{this, "DefaultPageSize",
                                               _("Default page size"), 5,
                                               IntConstrain(1, 10)};
+#ifdef ENABLE_KEYBOARD
```
wengxt, line 153: You can consider use HiddenOption for these two when enable_keyboard is false.

wengxt (changes requested):

Issue #430, opened 2022-01-12
Changing IM freezes Weston


daipom:
**Describe the bug**

I'm running Fcitx5 on Weston 5.0.0.
Since the commit of d711f4a92fc4c13d1c498b74430165f341e7f6ac, Weston freezes as follows.

When I re-focus a text input area after `ZwpInputPanelSurfaceV1::setOverlayPanel` in `WaylandInputWindow::update` is called, Weston freezes.

**To Reproduce**

1. Run Weston 5.0.0 with Fcitx5 (since the commit of d711f4a92fc4c13d1c498b74430165f341e7f6ac) as input-method.
1. Run Weston-Editor on Weston.
1. Focus a text input area on Weston-Editor.
1. Change IM by `Ctrl+Space` key. This calls `ZwpInputPanelSurfaceV1::setOverlayPanel` in `WaylandInputWindow::update`.
1. Unfocus and re-focus a text input area on Weston-Editor (Click an empty area and click the input area).
1. Weston freezes.

* Fcitx5 profile

```ini
[Groups/0]
# Group Name
Name=default
# Layout
Default Layout=us
# Default Input Method
DefaultIM=anthy

[Groups/0/Items/0]
# Name
Name=keyboard-us
# Layout
Layout=

[Groups/0/Items/1]
# Name
Name=anthy
# Layout
Layout=

[GroupOrder]
0=default
```

* Weston.ini

```ini
[core]
idle-time=0
xwayland=true

[keyboard]
keymap_rules=evdev
keymap_layout=jp

[output]
name=X1
mode=1024x768

[output]
name=wayland0
mode=1024x768

[input-method]
path=/home/daipom/bin/fcitx-weston.sh

[launcher]
path=/opt/weston/bin/weston-terminal
icon=/opt/weston/share/weston/icon_terminal.png

[launcher]
path=/opt/weston/bin/weston-editor
icon=/opt/weston/share/weston/icon_editor.png
```

* How to run Fcitx5 on Weston
  * I build Fcitx5 into `/opt/fcitx5` and use a shell script as follows to run it on Weston.
  * This script is set to `input-method` in `Weston.ini`.

```sh
#!/bin/sh
export DISPLAY=:1
export LD_LIBRARY_PATH="/opt/fcitx5/lib:$LD_LIBRARY_PATH"
/opt/fcitx5/bin/fcitx5 --verbose "*=5"
```

* Weston editor
[...]

PR #431, opened 2022-01-12
Fix group enumerating

daipom:
For simplification, the following describes the case where `forwarding` is `true`.

There is the problem where `Instance::enumerateGroup` and `InstancePrivate::navigateGroup` can't select the third and later groups.

In `navigateGroup`, `inputState->pendingGroupIndex_` is always `1` because every time `acceptGroupChange` is called, it goes back to `0`.
`InputMethodManager::setCurrentGroup` swaps the currently selected group with the newly selected one.
So the current group enumerating logic just swaps the first group and the second group, and can't select the third and later groups.

This PR fixes this problem.

I have deleted the notification logic in `navigateGroup`.
This is because this notification is executed at the following logic after all.

https://github.com/fcitx/fcitx5/blob/e537bc9e2f6158a0873656a4f6d4be1230cff850/src/lib/fcitx/instance.cpp#L796-L818

Note: Changing groups freeze my environment because of #430, so I have tested this fix with the following fix.

```diff
--- a/src/ui/classic/waylandinputwindow.cpp
+++ b/src/ui/classic/waylandinputwindow.cpp
@@ -115,7 +115,7 @@ void WaylandInputWindow::update(fcitx::InputContext *ic) {
             panelSurfaceV2_.reset(im->getInputPopupSurface(window_->surface()));
         }
     } else if (ic->frontend() == std::string_view("wayland")) {
-        panelSurface_->setOverlayPanel();
+        // panelSurface_->setOverlayPanel();
     }
     if (!panelSurface_ && !panelSurfaceV2_) {
         return;
```

### DIFF
--- a/src/lib/fcitx/inputmethodmanager.cpp
+++ b/src/lib/fcitx/inputmethodmanager.cpp
@@ -276,6 +276,28 @@ const InputMethodGroup &InputMethodManager::currentGroup() const {
     return d->groups_.find(d->groupOrder_.front())->second;
 }
 
+void InputMethodManager::enumerateGroup(bool forward) {
+    FCITX_D();
+    if (groupCount() < 2) {
+        return;
+    }
+    emit<InputMethodManager::CurrentGroupAboutToChange>(d->groupOrder_.front());
+    if (forward) {
+        d->groupOrder_.splice(
+            d->groupOrder_.end(),
+            d->groupOrder_,
+            d->groupOrder_.begin()
+        );
+    } else {
+        d->groupOrder_.splice(
+            d->groupOrder_.begin(),
+            d->groupOrder_,
+            --(d->groupOrder_.end())
+        );
+    }
+    emit<InputMethodManager::CurrentGroupChanged>(d->groupOrder_.front());
+}
+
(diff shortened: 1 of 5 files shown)


wengxt:
No, that's actually intended behavior. You missed the part of "totally released".

For example, with the default key "super+space", if you hold super, and continuously press space, you'll be able to move to the 3rd/4th group.

The logic here is to implement an Alt+Tab stacking style switching between different groups. The reason behind this is to keep sync with other platform e.g. windows/mac/ibus. So if user prefer the global input method switching style, they can choose to use 1 IM per group to get the similar behavior.

Changing it to rotation is not what I want. Also, there's a reason behind group changing requires two phase (enumerate -> accept). For the XCB implementation, key grab should be released before changing the layout, otherwise user may set a hotkey for example Super + Z, and super + Z may become a different key (e.g. qwertz's Y) under different layout, this will prevent user to continuously press the same key to switch layout.

daipom:
Thank you for the details!
I missed the fact that I can select the third and subsequent ones by holding the modifier key.
I understand the intent of this implementation.
Thank you!

PR #436, opened 2022-01-19
Enable KeyboardEngine to use hint by default

daipom:
Currently, we have to input `Control+Alt+H` to enable hint in KeyboardEngine.
I think it would be more useful if we could use hint by default in KeyboardEngine.

This fix adds `EnableHintByDefault` parameter to keyboard.conf.
By setting this parameter to `True`, we can use hint by default in KeyboardEngine.

~/.config/fcitx5/conf/keyboard.conf
```ini
EnableHintByDefault=True
```

I have confirmed this works correctly in keyboard-us.

### DIFF
--- a/src/im/keyboard/keyboard.cpp
+++ b/src/im/keyboard/keyboard.cpp
@@ -327,6 +327,16 @@ std::vector<InputMethodEntry> KeyboardEngine::listInputMethods() {
     return result;
 }
 
+void KeyboardEngine::activate(const InputMethodEntry &entry,
+                              InputContextEvent &event) {
+    if (config_.enableHintByDefault.value() &&
+        supportHint(entry.languageCode())) {
+        auto *inputContext = event.inputContext();
+        auto *state = inputContext->propertyFor(&factory_);
+        state->enableWordHint_ = true;
+    }
+}
+
 void KeyboardEngine::reloadConfig() {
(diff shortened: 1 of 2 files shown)


src/im/keyboard/keyboard.cpp
```diff
@@ -327,6 +327,16 @@ std::vector<InputMethodEntry> KeyboardEngine::listInputMethods() {
     return result;
 }
 
+void KeyboardEngine::activate(const InputMethodEntry &entry,
```
wengxt, line 330: I think this should be done in the state constructor.

Engine::activate is basically focus in.

wengxt (changes requested):

src/im/keyboard/keyboard.cpp
daipom: Thank you for the advice.
I have fixed this.

wengxt:
Thanks! I've been wondering if you guys are developing a solution for virtual keyboard based on fcitx5, is that the case?

daipom:
Yes, we are developing a virtual keyboard based on fcitx5.
One of our customers needs a virtual keyboard, so we're developing it as a POC for now.
We have not yet considered the consistency of the implementation at this moment,
but we would like to do so eventually.

daipom:
Thanks as always for the review and the merge!

PR #441, opened 2022-01-28
Add log rule notimedate

berberman:


src/lib/fcitx-utils/log.cpp
```diff
@@ -23,7 +23,10 @@ namespace {
 
 FCITX_DEFINE_LOG_CATEGORY(defaultCategory, "default");
 
+using log_rules_t = std::vector<std::pair<std::string, LogLevel>>;
```
wengxt, line 26: Please use CamelCase for type name.

Also, I suggest you create a alias for std::pair<std::string, LogLevel>, not the vector type. It would be more readable for knowing that this is a vector.

src/lib/fcitx-utils/log.cpp
```diff
@@ -50,26 +53,12 @@ class LogRegistry {
         categories_.erase(&category);
     }
 
-    void setLogRule(const std::string &ruleString) {
+    void setLogRule(const log_rules_t &rules) {
         std::lock_guard<std::mutex> lock(mutex_);
 
         rules_.clear();
```
wengxt, line 59: I think you can just do rules_ = rules here.

src/lib/fcitx-utils/log.cpp
```diff
@@ -164,7 +153,26 @@ bool LogCategory::fatalWrapper2(LogLevel level) {
 const LogCategory &Log::defaultCategory() { return fcitx::defaultCategory(); }
 
 void Log::setLogRule(const std::string &ruleString) {
-    LogRegistry::instance().setLogRule(ruleString);
+    log_rules_t parsedRules;
+    auto rules = stringutils::split(ruleString, ",");
+    for (const auto &rule : rules) {
+        auto ruleItem = stringutils::split(rule, "=");
+        if (ruleItem.size() != 2) {
+            if (ruleItem.size() == 1 && ruleItem[0] == "notimedate")
```
wengxt, line 161: Suggest check special flag before split with "=". So if we want to have some more special values, it can be put together.

if (rule == "'...") {
     ....
} else if (rule == "...") {
    ....
} else {
}

src/lib/fcitx-utils/log.cpp
```diff
@@ -23,7 +23,10 @@ namespace {
 
 FCITX_DEFINE_LOG_CATEGORY(defaultCategory, "default");
 
+using log_rules_t = std::vector<std::pair<std::string, LogLevel>>;
+
 static std::ostream *defaultLogStream = &std::cerr;
```
wengxt, line 28: Since we have more than one global value, suggest to create a new struct to hold them all and have only one top level global variable.

wengxt (changes requested):

PR #447, opened 2022-02-13
Add X-Cinnamon Desktop Environment Variable

xingwozhonghua:
Add X-Cinnamon Desktop Environment Variable


wengxt:
It's not really useful :), because we have no special code path for cinnamon anyway.

Issue #87, opened 2020-07-28
fcitx5 支援 Virtual Key Board？


GO-jack:
您好
想请问 fcitx5 是否支援 Virtual Key Board 或未来是否有计画支援？
谢谢

wengxt:
不会支持 fcitx 4 的那种，只会考虑正常触摸屏的那种。

wengxt:
leave it open for real virtual keyboard support.

SamLukeYes:
> 不会支持 fcitx 4 的那种，只会考虑正常触摸屏的那种。

很期待这个功能 :)

xingwozhonghua:
https://github.com/linuxmint/onboard

fcitx5后面还有什么任务可以适合新人参与那？

wengxt:
https://docs.google.com/document/d/1x-UcMQLLDi-ksxCQgb3L9AAuXGUYepNGkPSDUm1VlXE/edit?usp=sharing

Bob0111:
> https://docs.google.com/document/d/1x-UcMQLLDi-ksxCQgb3L9AAuXGUYepNGkPSDUm1VlXE/edit?usp=sharing

参考实现可以参考Ubuntu touch、kde plasma Mobile等Linux手机项目。这些项目已经在wayland下实现了虚拟键盘。你们可以考虑把fcitx作为一个附加模块加上去。

wengxt:
@daipom @ashie so as for the goal for supporting virtual keyboard in fcitx, the goal is that
1. have an abstraction for describing virtual keyboard layout
2. the abstraction can be displayed in a separate process, like kimpanel

Fcitx's own UI is kinda hacky with limited functionality, due to it doesn't use a tool kit.

There's currently another project that tries to bring fcitx to android, https://github.com/rocka/fcitx5-android-poc/, but they have the same issue like you guys: there's no abstraction, so the keyboard is also hard coded.

I'm not yet decided whether the desktop input method and the virtual keyboard should be managed in the same way, because of the existence of hybrid device.

As for display the VK window, I think it's OK to not use the wayland im protocol, since the virtual keyboard is kinda a fixed position if shell support protocol to place a window on the screen edge. 

Also, as for the abstraction, I'm thinking about to separate the key label and key shape, because re-create the key widget might be an expensive operation in certain cases. That leads to what is currently in the virtual keyboard branch, though very incomplete for now.

wengxt:
https://gitee.com/openkylin/kylin-virtual-keyboard

PR #450, opened 2022-02-16
Make Fcitx5::Module::WaylandIM usable from outside addons

daipom:
We are creating some addons, and we need to use `Fcitx5::Module::WaylandIM`.

Currently, we can use `Fcitx5::Module::Wayland`, but cannot use `Fcitx5::Module::WaylandIM`.
It is because there are no cmake files of `Fcitx5::Module::WaylandIM` such as `Fcitx5ModuleWaylandIMConfig.cmake`.

Is the reason there is no `INSTALL` to `fcitx5_export_module` of waylandim because it is not supposed to be used from an external addon?
If we have an addon needing to use this, is it possible to have such a modification put in?

### DIFF
--- a/src/frontend/waylandim/CMakeLists.txt
+++ b/src/frontend/waylandim/CMakeLists.txt
@@ -17,4 +17,4 @@ fcitx5_translate_desktop_file(${CMAKE_CURRENT_BINARY_DIR}/waylandim.conf.in wayl
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/waylandim.conf" DESTINATION "${FCITX_INSTALL_PKGDATADIR}/addon"
         COMPONENT config)
 
-fcitx5_export_module(WaylandIM TARGET waylandim BUILD_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}" HEADERS waylandim_public.h)
+fcitx5_export_module(WaylandIM TARGET waylandim BUILD_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}" HEADERS waylandim_public.h INSTALL)


wengxt:
The reason that it's not public is because there's no such use case. 
Also, fcitxwayland is not a public library. Current function returns a pointer of ZwpInputMethodV2. I'm not sure if you want to copy over the code or sth else.

wengxt:
Can you explain what you want to use with it and maybe we can find a better way to collaborate together.

E.g. may be we want to make FcitxWayland public, the code used to generate can also be made public if needed.

Or, another possibility is to use the wayland c interface to pass the object.

ashie:
> Also, fcitxwayland is not a public library. Current function returns a pointer of ZwpInputMethodV2. I'm not sure if you want to copy over the code or sth else.

Yes, we want you to export them, or want to discuss better way than it.
This pull request is the first step for it.
(But, sorry, now I've found that we might not need it, as describe it later.)

As you know, we are developing a virtual keyboard based on ClassicUI of Fcitx5:

* https://github.com/clear-code/fcitx5/tree/5.0.10+virtual-keyboard
  * Squashed version: https://github.com/clear-code/fcitx5/tree/5.0.10+virtual-keyboard-squashed
  * Based on master branch: https://github.com/clear-code/fcitx5/tree/virtual-keyboard
* https://www.youtube.com/watch?v=f1QBqWy_Ps4

(Our main target is Weston which has only input-method-v1 protocol, but it would be nice if it works also on X11 as well.)

We are now considering how to publish it.
Plan A is creating a pull request as is and try to merge it into Fcitx5 upstream.
But I think it's not appropriate for now, because most features are hardcoded yet and probably it doesn't follow your plan: https://github.com/fcitx/fcitx5/tree/virtualkeyboard

As Plan B, we are now trying to extract the code as a new UI addon:

* https://github.com/clear-code/fcitx5-virtualkeyboard-ui/

The base code of UI is copied from Classic UI and added our virtual keyboard code to it.
It's not buildable yet since wayland related APIs of Fcitx5 aren't exported.
Copying also fcitx-wayland might solve some of them but ...

... Now I've noticed that at least we don't need waylandim, because our target is only  input-method-v1.
When I remove v2 related code, I get succeeded to build it without exported waylandim.

* https://github.com/clear-code/fcitx5-virtualkeyboard-ui/tree/remove-input-method-v2

Now copying also fcitx-wayland might solve out problem completely.
Sorry for bothering you, we need more brushing up our plan before suggesting anything.

ashie:
I've succeeded to build [fcitx5-virtualkeyboard-ui/](https://github.com/clear-code/fcitx5-virtualkeyboard-ui/) by copying fcitx-wayland and removing input-method-v2 code.
Thanks for your hint!

Now this pull request can be closed for us, if no one else want it.

BTW we have some more changes in Fcitx5 core to implement our virtual keyboard code, so we'll create pull requests for them later.

daipom:
> Can you explain what you want to use with it and maybe we can find a better way to collaborate together.
> 
> E.g. may be we want to make FcitxWayland public, the code used to generate can also be made public if needed.
> 
> Or, another possibility is to use the wayland c interface to pass the object.

Thank you!

We are trying to do what @ashie says above.

As @ashie says, it turns out that we don't need `waylandim` after all,
so this pull request can be closed. I'm sorry for bothering you.

We would like to continue to create other pull requests to make a virtual keyboard as a UI module,
so please collaborate with us.

wengxt:
I'll close this for now, but feel free to leave comment on https://github.com/fcitx/fcitx5/issues/87

PR #452, opened 2022-02-25
Add simple group enumerating function for addons

daipom:
Our virtual keyboard addon needs to simply enumerate the current IM group.
There is a switching language button, and we would like to enumerate the current IM group by pressing it.

I've sent a similar PR before: https://github.com/fcitx/fcitx5/pull/431
and I understand Fcitx5 has its own stacking style switching function to control the current IM group.

But how about just adding a function of simple enumeration to `InputMethodManager` for addons?

### DIFF
--- a/src/lib/fcitx/inputmethodmanager.cpp
+++ b/src/lib/fcitx/inputmethodmanager.cpp
@@ -276,6 +276,28 @@ const InputMethodGroup &InputMethodManager::currentGroup() const {
     return d->groups_.find(d->groupOrder_.front())->second;
 }
 
+void InputMethodManager::enumerateGroup(bool forward) {
+    FCITX_D();
+    if (groupCount() < 2) {
+        return;
+    }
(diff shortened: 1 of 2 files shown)


src/lib/fcitx/inputmethodmanager.cpp
```diff
@@ -276,6 +276,28 @@ const InputMethodGroup &InputMethodManager::currentGroup() const {
...
+        d->groupOrder_.splice(
+            d->groupOrder_.end(),
+            d->groupOrder_,
+            d->groupOrder_.begin()
+        );
+    } else {
+        d->groupOrder_.splice(
+            d->groupOrder_.begin(),
+            d->groupOrder_,
+            --(d->groupOrder_.end())
```
wengxt, line 295: nit: I prefer to use std::prev here.

daipom: Thank you for your review!
I have fixed it.

daipom:
Thanks for merging!

PR #454, opened 2022-02-28
Add virtual key handling to InputContext and WaylandIMInputContextV1

daipom:
For a virtual keyboard, we need an interface to convey virtual key inputs to InputContext.

When we were beginning to implement the virtual keyboard, we tried to manage virtual key inputs
using `keyEvent` and `forwardKey` and `commitString` and so on...
It was tough to handle them properly on the virtual keyboard side.

It would be nice if we could pass virtual keys to the same logic as physical keys.

In `WaylandIMInputContextV1`, key inputs are processed in the `keyCallback` method,
but there is no way to pass virtual key inputs to it.

This fix adds new methods `virtualKeyEvent` and `virtualKeyEventImpl` to InputContext.

By default, it works the same as the existing `keyEvent` method,
but `virtualKeyEventImpl` can be overridden to handle virtual key events.

In this fix, core key logic can be shared between physical and virtual key events.

### DIFF
--- a/src/frontend/waylandim/waylandimserver.cpp
+++ b/src/frontend/waylandim/waylandimserver.cpp
@@ -427,13 +427,44 @@ void WaylandIMInputContextV1::keyCallback(uint32_t serial, uint32_t time,
                        server_->modifiers_, code),
                    state == WL_KEYBOARD_KEY_STATE_RELEASED, time);
 
-    if (state == WL_KEYBOARD_KEY_STATE_RELEASED && key == repeatKey_) {
+    processKeyEvent(event, false, serial);
+}
+
+void WaylandIMInputContextV1::virtualKeyEventImpl(KeyEvent &event) {
+    // `mods_locked` value seems to be 16 usually.
+    // If we need to implement behavior such as `capslocked` for virtual key events, 
+    // we will need to manage this `mods_locked` value.
+    if (event.rawKey().hasModifier()) {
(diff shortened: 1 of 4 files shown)


wengxt:
Hi, sorry for not replying this for a while. The key point for me is that, before we move forward we need to have a better design on how it should looks like.

For example, "KeyEvent" is solely designed for hardware keyboard. The property on it is not really suitable for virtual key.

While the change looks ok on its own, I don't think it fit the big picture, especially virtual key may do things that hardware key can't.

I'd suggest we find some time and meet together (maybe also with people that's currently working on fcitx5-android). What do you think?

Feel free to leave a message on irc or telegram in fcitx channel

Telegram: https://fcitx-im.org/telegram/captcha.html
IRC: [#fcitx [at] libera.chat](https://web.libera.chat/?channels=#fcitx)

daipom:
Thank you for the response!

I would appreciate it if there is a proper input method for the virtual keys,
but I too think it doesn't have to be this implementation.

We need to figure out in which direction Fcitx5 plans to support virtual key operations in the big picture.

> I'd suggest we find some time and meet together (maybe also with people that's currently working on fcitx5-android). What do you think?
> Feel free to leave a message on irc or telegram in fcitx channel

Thank you!
I don't have much time at the moment so action may be slow,
but I hope to contribute to the virtual keyboard feature of Fcitx5.

PR #457, opened 2022-03-07
extra “;" after member function definition

san-Hei:
extra “;" after member function definition


wengxt:
no, though not required, it looks better these way.

PR #474, opened 2022-04-06
add support of xcb input panel offset

Reverier-Xu:
Support for configuring the overall position offset of the InputPanel.

### DIFF
--- a/src/ui/classic/xcbinputwindow.cpp
+++ b/src/ui/classic/xcbinputwindow.cpp
@@ -91,9 +91,13 @@ void XCBInputWindow::updatePosition(InputContext *inputContext) {
         y = newY;
     }
 
+    // add support of input panel offset here.
+    auto &theme = parent_->theme();
+
     xcb_params_configure_window_t wc;
-    wc.x = x;
-    wc.y = y;
+    wc.x = x + theme.inputPanel->offsetX.value();
+    wc.y = y + theme.inputPanel->offsetY.value();
+
     wc.stack_mode = XCB_STACK_MODE_ABOVE;
     xcb_aux_configure_window(ui_->connection(), wid_,
                              XCB_CONFIG_WINDOW_STACK_MODE |
--- a/src/ui/classic/xcbinputwindow.cpp
+++ b/src/ui/classic/xcbinputwindow.cpp
@@ -56,8 +56,8 @@ void XCBInputWindow::updatePosition(InputContext *inputContext) {
     topSW = theme.inputPanel->shadowMargin->marginTop.value();
     bottomSW = theme.inputPanel->shadowMargin->marginBottom.value();
 
-    x = inputContext->cursorRect().left() - leftSW;
-    y = inputContext->cursorRect().top() - topSW;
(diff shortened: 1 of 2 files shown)


wengxt:
you should define 4 value instead of just two. (hint reuse the existing margin type)
1. fcitx window will be flipped on screen edge, you need to consider the bottom shadow when flipped. in the future when we support RTL, the right width of shadow might be used too.
2. fcitx window placement calculation will then need to exclude the shadow border, instead of using the raw window size.

Reverier-Xu:
But offset just moves the whole input window on the screen, do we really need to set 4 values and exclude shadow border?
Fcitx5 now have `ContentMargin` and `BackgroundMargin` to set space width around the content, shadow margins maybe duplicated.

Reverier-Xu:
Oh, I understand what you mean. When the input window is out of the screen range, it will be transferred to the top of the screen. The offset under the RTL layout needs to be reversed, so it needs 4 shadow widths. Thank you.

src/ui/classic/xcbinputwindow.cpp
```diff
@@ -47,10 +47,22 @@ void XCBInputWindow::updatePosition(InputContext *inputContext) {
...
+    leftSW = theme.inputPanel->shadowMargin->marginLeft.value();
+    rightSW = theme.inputPanel->shadowMargin->marginRight.value();
+    topSW = theme.inputPanel->shadowMargin->marginTop.value();
+    bottomSW = theme.inputPanel->shadowMargin->marginBottom.value();
+
+    x = inputContext->cursorRect().left() - leftSW;
+    y = inputContext->cursorRect().top() - topSW;
     h = inputContext->cursorRect().height();
 
+    expectedWidth = width() - leftSW - rightSW;
```
wengxt, line 63: expectedWidth is a little bit confusing, I'd prefer sth like actualWidth. Also, you need to make sure it's not less than zero.

src/ui/classic/xcbinputwindow.cpp
```diff
@@ -47,10 +47,22 @@ void XCBInputWindow::updatePosition(InputContext *inputContext) {
...
+    // add support of input panel offset here.
+    // TODO: RTL support.
+    auto &theme = parent_->theme();
+    int leftSW, rightSW, topSW, bottomSW, expectedWidth, expectedHeight;
+    leftSW = theme.inputPanel->shadowMargin->marginLeft.value();
+    rightSW = theme.inputPanel->shadowMargin->marginRight.value();
+    topSW = theme.inputPanel->shadowMargin->marginTop.value();
+    bottomSW = theme.inputPanel->shadowMargin->marginBottom.value();
+
+    x = inputContext->cursorRect().left() - leftSW;
```
wengxt, line 59: I don't think you should change "x" and "y" here. Basically what we want to do is that

we have a rect with size (actualW*actualH) be put at (x,y).

The x,y calculation should not consider the shadow offset.

After all the existing calculation, we move the window by (-leftOffset, -topOffset).

wengxt (changes requested):

PR #494, opened 2022-05-23
fix Back Forward -> Backward

chenshijie-uos:


wengxt:
No, the intention is to exact match the key name. Enum name comes from xkb.

PR #520, opened 2022-06-11
Remove trailing slash from dir paths before checking uniqueness. Fix #519

### LINKED ISSUE #519: Standardpath uses same directory more than once
### Steps to reproduce
Use ubuntu 22.04

> $ echo $XDG_DATA_DIRS
> /usr/share/ubuntu:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop

Trailing '/' is not a good practice, but this is what is given from this distribution.

In addons, use `standardPath::scandirectories(StandardPath::Type::Data, callback)`

### Expected
The callback is called by unique paths.

### Actual
The callback is called with `/usr/local/share` and `/usr/local/share/`, same in the sense of path.

naokiri:
Remove trailing slash from dir paths before checking uniqueness. This fixes fcitx#519

### DIFF
--- a/src/lib/fcitx-utils/standardpath.cpp
+++ b/src/lib/fcitx-utils/standardpath.cpp
@@ -199,6 +199,9 @@ private:
         }
 
         auto rawDirs = stringutils::split(dir, ":");
+        for (auto &rawDir : rawDirs) {
+            rawDir = fs::cleanPath(rawDir);
+        }
         std::unordered_set<std::string> uniqueDirs(rawDirs.begin(),
                                                    rawDirs.end());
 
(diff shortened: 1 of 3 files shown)


src/lib/fcitx-utils/fs.cpp
```diff
@@ -143,6 +143,14 @@ std::string cleanPath(const std::string &path) {
     return buf;
 }
 
+std::string cleanDirPath(const std::string &dirPath) {
+    auto cleanedPath = cleanPath(dirPath);
+    while (!cleanedPath.empty() && cleanedPath.back() == '/') {
```
wengxt, line 148: This would be wrong if dirPath contains only slash.

wengxt (changes requested):
Please also add test case to this function.

wengxt:
Actually I believe there's no need to use cleanDirPath. cleanPath can already remove trailing slash.

naokiri:
You're right the cleanPath already did the job. Thanks for fixing!

PR #529, opened 2022-06-22
当前deepin桌面环境下使用fcitx5点击配置跳转到控制中心输入法选项

chenshijie-uos:
当前deepin桌面环境下使用fcitx5点击配置跳转到控制中心输入法选项，

### DIFF
--- a/data/fcitx5-configtool.sh
+++ b/data/fcitx5-configtool.sh
@@ -65,6 +65,10 @@ detectDE() {
             XFCE)
             DE=xfce
             break
+            ;;
+            Deepin)
+            DE=deepin
+            break
         esac
       done
     fi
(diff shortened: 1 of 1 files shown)


data/fcitx5-configtool.sh
```diff
@@ -156,6 +167,9 @@ case "$DE" in
     kde)
         order="kde qt xdg"
         ;;
+    deepin)
+        order="dde"
```
wengxt, line 171: please fallback to non dde version. anything is better than nothing.

data/fcitx5-configtool.sh
```diff
@@ -125,6 +129,13 @@ run_qt() {
     return 1
 }
 
+run_dde() {
+    if which dde-control-center > /dev/null 2>&1; then
+        dbus-send --print-reply --dest=com.deepin.dde.ControlCenter /com/deepin/dde/ControlCenter com.deepin.dde.ControlCenter.ShowPage string:'keyboard/Manage Input Methods' > /dev/null 2>&1
```
wengxt, line 134: Is this page dedicated for fcitx? or it's for deepin in general? What does it looks like?

Also I think it better provides a way to detect such page exists or not.

If dde-control-center can have a flag to be used in a similar way like systemsettings5 above, I'd prefer to use that instead of just dbus-send.

wengxt (changes requested):

PR #550, opened 2022-07-19
Proposal: add option to change wayland ui resolution

omegacoleman:
There's currently DPI detection logic in XCB ui, but wayland applications need to handle DPI themselves, and this option is currently missing. On my 1080p screen, the wayland ui candidate box is too small, as it rendered in the default 96 dpi. I could workaround it by setting a huge font size, but that would also increase the candidate box size in qt & gtk.

This PR add an option to classicui configure section, setting a DPI separately for wayland ui.

### DIFF
--- a/src/ui/classic/classicui.h
+++ b/src/ui/classic/classicui.h
@@ -132,7 +132,22 @@ FCITX_CONFIGURATION(
                "configuration needs to support this to use this feature.")}};
     Option<std::string, NotEmpty, DefaultMarshaller<std::string>,
            ThemeAnnotation>
-        theme{this, "Theme", _("Theme"), "default"};);
+        theme{this, "Theme", _("Theme"), "default"};
+    Option<int, IntConstrain, DefaultMarshaller<int>, ToolTipAnnotation>
+        forceWaylandDPI{
+            this,
+            "ForceWaylandDPI",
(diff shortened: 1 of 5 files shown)


CoelacanthusHex:
You might be better off formatting the code, I've found some mixed tabs and spaces.

omegacoleman:
> You might be better off formatting the code, I've found some mixed tabs and spaces.

done, forgot that

wengxt:
The request itself sounds weird to me. If the window under wayland shows with a wrong scale, the it might just be scale for screen is wrong? can you just set your screen scale to something like 125%?

omegacoleman:
> The request itself sounds weird to me. If the window under wayland shows with a wrong scale, the it might just be scale for screen is wrong? can you just set your screen scale to something like 125%?

That would cause Xwayland apps to be blurry, btw not even all native wayland apps support fractional scaling well. To make everything nice & sharp, we prefer toolkit layer (renderer) font DPI settings, like QT_FONT_DPI and dconf text-scaling-factor

In short, the one renders the text shall handle font DPI properly. fcitx5-qt and gtk already work well this way.

scaling is like X11 xrandr scaling opts, while text DPI support corresponds to Xft dpi settings, just like what implemented in xcb now. They may serve a common purpose (hidpi) but were indeed different things.

PR #563, opened 2022-08-08
Add virtual keyboard in fcitx5 framework

kingysu:


src/frontend/fcitx4frontend/fcitx4frontend.cpp
```diff
@@ -253,6 +256,24 @@ class Fcitx4InputContext : public InputContext,
         return keyEvent(event) ? 1 : 0;
     }
 
+    bool isVirtualKeyboardVisibleDBus() const {
```
wengxt, line 259: there should be no addition to fcitx4 frontend.

src/lib/fcitx/instance.h
```diff
@@ -73,6 +73,13 @@ enum class EventWatcherPhase {
 
 struct FCITXCORE_EXPORT InstanceQuietQuit : public std::exception {};
 
+enum class DeviceMode {
+
+    PC = 0,
```
wengxt, line 78: naming is a little bit confusing here.

src/modules/devicemodewatcher/devicemodewatcher.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * SPDX-FileCopyrightText: 2017-2017 CSSlayer <wengxt@gmail.com>
```
wengxt, line 2: Why use my copyright?

src/modules/devicemodewatcher/devicemodewatcher.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "fcitx/addonfactory.h"
+#include "devicemodewatcher.h"
+
+namespace fcitx {
+
+DeviceModeWatcher::DeviceModeWatcher(Instance *instance)
+    : instance_(instance), bus_(dbus()->call<IDBusModule::bus>()),
+      watcher_(*bus_) {
+    entry_ = watcher_.watchService(
+        "com.kylin.statusmanager.interface",
```
wengxt, line 17: Please put all the new addon in to separate PR.

src/lib/fcitx/instance.h
wengxt: Also what's the difference between device mode and input method mode? why it needs to be two different enum?

src/lib/fcitx/userinterface.cpp
```diff
@@ -10,4 +10,14 @@
 namespace fcitx {
 
 UserInterface::~UserInterface() {}
+
+bool UserInterface::isVirtualKeyboard() {
+    return false;
+}
+
+VkUserInterface::~VkUserInterface() {}
```
wengxt, line 18: Use full name instead of "vk".

src/lib/fcitx/userinterface.h
```diff
@@ -46,6 +46,20 @@ class FCITXCORE_EXPORT UserInterface : public AddonInstance {
     virtual bool available() = 0;
     virtual void suspend() = 0;
     virtual void resume() = 0;
+    virtual bool isVirtualKeyboard();
```
wengxt, line 49: I think this should be bool isVirtualKeyboard const(). But put that aside, if no addon will support both at the same time, I'd suggest add a new field in AddonInfo instead. So we know before the loading happens.

kingysu: In your opinion, what's the right class to add the new field? class AddonInfoPrivate or class AddonConfig？ And what's the name of the new field?

src/lib/fcitx/userinterface.cpp
kingysu: I will replace "vk" with "VirtualKeyboard"

src/modules/devicemodewatcher/devicemodewatcher.cpp
kingysu: I will seperate the new addon into new PR.

src/modules/devicemodewatcher/devicemodewatcher.h
kingysu: It's a typo. Sorry.

src/lib/fcitx/instance.h
kingysu: Input method mode is used to differentiate virtual keyboard and physical keyboard.
Device mode is used to differentiate if the device is in PC mode or Tablet mode.

src/frontend/fcitx4frontend/fcitx4frontend.cpp
kingysu: Could you explain more?

src/lib/fcitx/instance.h
kingysu: So far, device mode is only used to not hide virtual keyboard when focus is out on PC if there is an available physical keyboard.
If there is no available physical keyboard or device mode is tablet, the virtual keyboard is automatically closed when focus is out.

src/frontend/dbusfrontend/dbusfrontend.cpp
```diff
@@ -282,6 +282,9 @@ class DBusInputContext1 : public InputContext,
     bool processKeyEvent(uint32_t keyval, uint32_t keycode, uint32_t state,
                          bool isRelease, uint32_t time) {
         CHECK_SENDER_OR_RETURN false;
+
+        im_->instance()->setInputMethodMode(InputMethodMode::PhysicalKeyboard);
```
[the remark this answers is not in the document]
kingysu: If the user show the virtual keyboard by dbus, the input method mode will be OnScreenKeyboard. Then, if the user press any physical keyboard key, we want the virtual keyboard to hide automatically. That's why I should change the input method mode in this function.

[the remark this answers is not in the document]
kingysu: Instead, you may want to use watchEvent on KeyEvent.——Could you explain more?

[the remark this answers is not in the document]
kingysu: Do you mean that I should watch the event: EventType::InputContextKeyEvent, so that there is no need to duplicate code in dbusfrontend and fcitx4frontend. If so, I will move the code before into the following codes:
`d->eventWatchers_.emplace_back(
        watchEvent(EventType::InputContextKeyEvent,
                   EventWatcherPhase::InputMethod, [this](Event &event) {
                       auto &keyEvent = static_cast<KeyEvent &>(event);
                       auto *ic = keyEvent.inputContext();
                       auto *engine = inputMethodEngine(ic);
                       const auto *entry = inputMethodEntry(ic);
                       if (!engine || !entry) {
                           return;
                       }
                       engine->keyEvent(*entry, keyEvent);
                   }));`

src/frontend/fcitx4frontend/fcitx4frontend.cpp
wengxt: There is no point to change fcitx 4 frontend, fcitx4 is dead code. You can't add new feature to fcitx4.

src/lib/fcitx/userinterface.h
wengxt: See how we have a field dedicated for UI Addon: https://github.com/fcitx/fcitx5/blob/08e0a5f9367466fb4360d0a8c2dbc80e9406d9ff/src/lib/fcitx/addoninfo.cpp#L26

You can add what ever you think that makes sense. for me, in order to allow flexibility, maybe a enum type "UIType" instead of bool.

src/frontend/dbusfrontend/dbusfrontend.cpp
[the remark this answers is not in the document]
wengxt: Yeah, and put that in a separate PR.

src/frontend/fcitx4frontend/fcitx4frontend.cpp
kingysu: > There is no point to change fcitx 4 frontend, fcitx4 is dead code. You can't add new feature to fcitx4.

So I should remove all the new code in fcitx4 frontend?

src/frontend/dbusfrontend/dbusfrontend.cpp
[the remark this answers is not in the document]
kingysu: > Yeah, and put that in a separate PR.

OK. It will be a very small PR.

PR #565, opened 2022-08-09
Add module: devicemodewatcher

kingysu:


wengxt:
Suggest to merge this in to the other addon. Seems the functionality are related.

(Adding 1 new addon instead of 2).

kingysu:
> Suggest to merge this in to the other addon. Seems the functionality are related.
> 
> (Adding 1 new addon instead of 2).

OK. If there is no need for device mode, the devicemodewatcher can be removed for now.

PR #611, opened 2022-10-04
fix fcitx5-diagnose requests environment variables to be set to fcitx…

marguerite:
… instead of fcitx5

### DIFF
--- a/data/fcitx5-diagnose.sh
+++ b/data/fcitx5-diagnose.sh
@@ -1113,18 +1113,18 @@ _env_incorrect() {
 
 check_xim() {
     write_title 2 "Xim:"
-    xim_name=fcitx
+    xim_name=fcitx5
     write_order_list "$(code_inline '${XMODIFIERS}'):"
     if [ -z "${XMODIFIERS}" ]; then
         write_error_eval "$(_ 'XMODIFIERS is not set')"
-        set_env_link XMODIFIERS '@im=fcitx'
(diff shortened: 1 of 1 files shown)


wengxt:
this is intended.

marguerite:
intended to use XMODIFIERS=@im=fcitx, GTK_IM_MODULE=fcitx and QT_IM_MODULE=fcitx? does fcitx5 recognize it?

wengxt:
recognize it since the very first day

marguerite:
does fcitx5 recognize both fcitx and fcitx5? I mean, if so, I can add something to the script to make it not complain and mislead

naokiri:
What made you think it should be "fcitx5" instead of "fcitx"?
As far as I know fcitx5 never expected it to be "fcitx5".
Also even for distribution's modification, AFAIK Ubuntu 22.04 never used or expected "fcitx5" as those modifiers. Maybe there's some other distribution did their own modification, then you should report a PR on their packaging.

marguerite:
I am the openSUSE packager, I used XMODIFIERS=@im=fcitx5 for a long time and my input method works like a charm...I even implement some systemd generators to distinguish fcitx and fcitx5 during start (I need to know if the user still uses fcitx to prompt a reminder for migration)...so I think that MODIFIER may also work since the very first day.

And if that's a mistake, I need to tweak my codes. so, you know, I need a guaranteed answer :-D

marguerite:
okay, I'll steadily move back to fcitx.

PR #632, opened 2022-10-29
Control the behavior of virtual keyboard when the object gets focus or loses focus

kingysu:
1. In some GUI frameworks like Qt, it provides good support for virtual keyboard.
    So Qt IM module has the need to disable the default behavior of virtual keyboard
    when the object get focus or loses focus. Qt can decide if it should show virtual
    keyboard when the object gets focus or hide virtual keyboard when the object loses
    focus.
    
    2. However, In some GUI frameworks like GTK, it provides poor support for virtual
    keyboard. So It's necessary to show or hide virtual keyboard when the object get focus
    or loses focus and this is the default behavior provided by fcitx5. Maybe in the future,
    GTK may provide good support for virtual keyboard just as Qt does.

### DIFF
--- a/src/frontend/dbusfrontend/dbusfrontend.cpp
+++ b/src/frontend/dbusfrontend/dbusfrontend.cpp
@@ -57,6 +57,15 @@ buildFormattedTextVector(const Text &text) {
     return vector;
 }
 
+std::string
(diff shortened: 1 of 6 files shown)


wengxt:
@kingysu 函数名太啰嗦了

PR #712, opened 2023-02-05
Implement Wayland xcb layout options

### LINKED ISSUE #705: Manage xkb options in addition to layout and variant
**Describe the feature**

I want fcitx5 to update the xkb options when switching groups. It properly switches the xkb layout and variant, but xkb options is right next to them and is neglected by fcitx5.

It is surprising that things were working when fcitx5 did not connect to Wayland and I managed layouts using KDE's keyboard layout switcher in plasmashell.  I had "lv3:ralt_switch_multikey" enabled for my us layout, and French accents were working. In my custom jp symbols file, I tried to undo the lv3 ralt so that my Henkan would work, and it did work for some reason. There might have been cases where my keyboard would be weird if I logged out without switching back to the us layout, but that would require me to go debug. The `setxkbmap -v 10` behavior is different between KDE's control and fcitx5's control. With KDE, it is strange that the setxkbmap does not change when switching layouts, while in fcitx5 the setxkbmap updates to display the current layout. That might have been why it was working for me before. An alternative solution would be to tell me how to get a layout variant file to have precedence over and override the level3 file. However, I think the proper way would be to complete fcitx5 to manage the xkb options.

**To Reproduce**
1. Have a us layout where I want `Options=lv3:ralt_switch_multikey` so that RALT lets me type French accents
2. Have a custom jp layout where I want `Options=` (empty options) so that RALT becomes Henkan (I don't like space or arrow keys, and it was working before fcitx) so that I can choose words
3. Observe that there is no way to get fcitx5 to change the xkb options to automatically support my setup.

**Expected behavior**

I should be able to configure the xkb options for each input group. It should support setting the options to empty. It should support options like "lv3:ralt_switch_multikey", but it should support other users' options as well and not hardcode this. Whenever fcitx5 switches to a new xkb layout and variant, it should ensure that the newly configurable xkb options are switched as well.  

**Analysis**

`Options=something` should be written by fcitx5, possible somewhere here: https://github.com/fcitx/fcitx5/blob/33ef1c10a4b4b7f4abd001703d1b50aecfec75b3/src/modules/wayland/waylandmodule.cpp#L212-L219

New keys should be added to `~/.config/fcitx5/profile`. "Default Layout Options" should be added next to "Default Layout" under "Groups/0". "Layout Options" should be added next to "Groups/0/Items/0". The options written to `kxkbrc` should match the values for these keys, including writing an empty value there for empty values here.

There does not need to be a config GUI in the first version. This is an advanced feature, and advanced users are fine editing the file. "System Settings > Input Devices > Keyboard > Advanced > Key to choose the 3rd level" is a somewhat complicated GUI to implement, so I was thinking some time and effort could be saved by leaving out the config GUI.

**Desktop (please complete the following information):**

This is a continuation of #704.

danielzgtg:
Closes #705.

This feature allows uses to have fcitx5 manage their KDE Wayland XKB layout options when switching groups.

# Disadvantages

- X11 is not implemented as it is more complicated
- Migration from kxkbrc is not implemented, and users may be shocked that their layout options are reset when they update fcitx, and be confused until they find this feature

# Before

I wanted to have a French Dvorak layout to type accents with RALT/AltGr level3, and a Japanese Kana layout with RALT mapped to Henkan.
 
I could have fcitx5 have one of them working automatically when switching between them. I had to change the "3rd level shortcuts" in "KDE System Settings > Input Devices > Keyboards > Layouts." When level3 was enabled, the French worked but the Japanese didn't as the level3 option overrode my custom definition of RALT as Henkan. When level3 was disabled, the French obviously didn't work as it was not enabled, but my custom Japanese RALT Henkan worked as the level3 options was no longer overriding it.

 # After

I put `Default Layout Options=lv3:ralt_switch_multikey` in my French group in `~/.config/fcitx5/profile`, and `Default Layout Options=` in my Japanese group. Now fcitx5 automatically controls my xcb keyboard layout so I can switch between languages conveniently.

# After

I can now

### DIFF
--- a/src/lib/fcitx/inputmethodconfig_p.h
+++ b/src/lib/fcitx/inputmethodconfig_p.h
@@ -23,6 +23,7 @@ FCITX_CONFIGURATION(
     Option<std::vector<InputMethodGroupItemConfig>> items{this, "Items",
                                                           "Items"};
     Option<std::string> defaultLayout{this, "Default Layout", "Layout"};
+    Option<std::string> defaultLayoutOptions{this, "Default Layout Options", "XKB Layout Options"};
     Option<std::string> defaultInputMethod{this, "DefaultIM",
                                            "Default Input Method"};);
 
--- a/src/lib/fcitx/inputmethodgroup.cpp
+++ b/src/lib/fcitx/inputmethodgroup.cpp
@@ -32,6 +32,7 @@ public:
     std::vector<InputMethodGroupItem> inputMethodList_;
     std::string defaultInputMethod_;
     std::string defaultLayout_;
+    std::string defaultLayoutOptions_;
 };
 
 InputMethodGroupItem::InputMethodGroupItem(const std::string &name)
@@ -131,5 +132,16 @@ void InputMethodGroup::setDefaultLayout(const std::string &layout) {
 const std::string &InputMethodGroup::defaultLayout() const {
     FCITX_D();
     return d->defaultLayout_;
+}
+
+void InputMethodGroup::setDefaultLayoutOptions(
+    const std::string &layoutOptions) {
+    FCITX_D();
+    d->defaultLayoutOptions_ = layoutOptions;
+}
+
+const std::string &InputMethodGroup::defaultLayoutOptions() const {
+    FCITX_D();
+    return d->defaultLayoutOptions_;
 }
 } // namespace fcitx
--- a/src/lib/fcitx/inputmethodgroup.h
+++ b/src/lib/fcitx/inputmethodgroup.h
@@ -45,6 +45,8 @@ public:
     const std::string &name() const;
     void setDefaultLayout(const std::string &layout);
     const std::string &defaultLayout() const;
+    void setDefaultLayoutOptions(const std::string &layoutOptions);
+    const std::string &defaultLayoutOptions() const;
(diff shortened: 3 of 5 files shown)


wengxt:
This is a behavior change. So I like to see an global option like "use system xkb option".

Actually, with this change, you may rename this option to "use system xkb option" https://github.com/fcitx/fcitx5/blob/master/src/lib/fcitx/globalconfig.cpp#L162 and remove this option: https://github.com/fcitx/fcitx5/blob/master/src/lib/fcitx/globalconfig.cpp#L173

Also, will you also work on expose this in the UI?

https://github.com/fcitx/fcitx5-configtool/

You may need to update https://github.com/fcitx/fcitx5/blob/46e051837a4ae7beb15a342adba18aedd00ae642/src/modules/dbus/dbusmodule.cpp#L172 too to allow this new setting to be sent over the dbus.

danielzgtg:
Thank you for the feedback. I will make sure to have that global option in the final version.

The last thing I was doing was reading the code. Before that I was installing the dev packages to get fcitx5-configtool to finally build. I need to fix a few strange problems of things randomly changing, and I need to understand the project to do that. I think this will take around 1 month, as I have university. Nobody else who might require this feature voted on GitHub, so I think timing shouldn't be a problem.

PR #744, opened 2023-03-16
Use a seperate service to switch input mode

kingysu:


wengxt:
It's ok if you want to use a separate object, but I see no point to use a separate service name.

src/ui/virtualkeyboard/virtualkeyboard.cpp
```diff
@@ -272,6 +272,13 @@ void VirtualKeyboard::updateInputPanel(InputContext *inputContext) {
     updateCandidate(inputContext);
 }
 
+void VirtualKeyboard::initVirtualKeyboardService() {
+    service_ = std::make_unique<VirtualKeyboardService>(this);
+    bus_->addObjectVTable("/virtualkeyboard",
+                          "org.fcitx.virtualkeyboard.service", *service_);
```
wengxt, line 278: We use interface name like [lower case reversedomain].Fcitx.CamelCaseServiceNameVersion 
So ideally should be called org.fcitx.Fcitx.VirtualKeyboard1

wengxt (changes requested):

src/ui/virtualkeyboard/virtualkeyboard.cpp
kingysu: ok

PR #769, opened 2023-04-13
Add a new dbus method to get current im list

kingysu:


wengxt:
Just use controller1 instead. If it doesn't return all the information you want, define a new function there.

kingysu:
> Just use controller1 instead. If it doesn't return all the information you want, define a new function there.

done

PR #780, opened 2023-04-26
Make the update logic of user interface more clear

kingysu:


wengxt:
I understood your intention, though I don't think you make it any better XD...

PR #784, opened 2023-05-04
 Optimize the support for virtual keyboard

kingysu:
1. There are two kinds of virtual keyboard at least.
    
2. The input method engine may support one or two function mode: full or limited.
   a) The full function mode is the default function mode. All the input method engine
   which does not implement InputMethodEngineV4 or its ascendants supports the full
   function mode.
   b) The limited function mode is a dedicated function mode which is used for the true
   virtual keyboard. The input method engine which implements InputMethodEngineV4 or its
   ascendants should tell fcitx which function mode it supports. That is, it may support
   full function mode or limited function mode or both.
    
3. The two different virtual keyboard use different function mode of input method engine.
   The virtual keyboard should tell fcitx which function mode it needs. Fcitx will decide
   which function mode is used. For example, if the virtual keyboard tells fcitx that it
   needs limited function mode, but the input method engine does not support that mode,
   then fcitx will use full function mode for compitibility reason.

### DIFF
--- a/src/ui/virtualkeyboard/virtualkeyboard.cpp
+++ b/src/ui/virtualkeyboard/virtualkeyboard.cpp
@@ -130,7 +130,7 @@ void VirtualKeyboardBackend::processKeyEvent(uint32_t keyval, uint32_t keycode,
         return;
     }
 
-    VirtualKeyboardEvent event(inputContext, time);
+    VirtualKeyboardEvent event(inputContext, isRelease, time);
     event.setKey(Key(static_cast<KeySym>(keyval), KeyStates(state), keycode));
 
     auto eventConsumed = inputContext->virtualKeyboardEvent(event);
--- a/src/ui/virtualkeyboard/virtualkeyboard.cpp
+++ b/src/ui/virtualkeyboard/virtualkeyboard.cpp
@@ -74,6 +74,8 @@ public:
 
     ~VirtualKeyboardBackend() = default;
 
+    void setVirtualKeyboardFunctionMode(uint32_t mode);
+
     void processKeyEvent(uint32_t keyval, uint32_t keycode, uint32_t state,
                          bool isRelease, uint32_t time);
 
@@ -106,6 +108,9 @@ private:
     PageableCandidateList *getPageableCandidateList();
 
 private:
+    FCITX_OBJECT_VTABLE_METHOD(setVirtualKeyboardFunctionMode,
+                               "SetVirtualKeyboardFunctionMode", "u", "");
+
     FCITX_OBJECT_VTABLE_METHOD(processKeyEvent, "ProcessKeyEvent", "uuubu", "");
 
     FCITX_OBJECT_VTABLE_METHOD(processVisibilityEvent, "ProcessVisibilityEvent",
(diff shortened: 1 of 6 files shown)


wengxt:
Question, should function mode be a capability flag instead ? if capability flag bits are limited, we can introduce a new property for input context. Associate it to every key event seems wrong to me.

kingysu:
> Question, should function mode be a capability flag instead ? if capability flag bits are limited, we can introduce a new property for input context. Associate it to every key event seems wrong to me.

I have updated the PR to make sure that the virtual keyboard can nofity fcitx5 of its function mode.

wengxt (changes requested):

src/lib/fcitx/inputcontext_p.h
```diff
@@ -132,6 +133,14 @@ class InputContextPrivate : public QPtrHolder<InputContext> {
 
     InputContextProperty *property(int slot) { return properties_[slot].get(); }
 
+    bool isFunctionModeSupported(FunctionMode functionMode) {
+        FCITX_Q();
+
+        auto *instance = manager_.instance();
+        auto *engine = instance->inputMethodEngine(q);
```
wengxt, line 140: engine can be null, even instance can be null.

src/lib/fcitx/inputmethodengine.cpp
```diff
@@ -28,18 +27,43 @@ std::string InputMethodEngine::subModeLabel(const InputMethodEntry &entry,
...
+        FCITX_WARN() << "InputMethodEngine::virtualKeyboardEvent is called."
+                     << "But there is no valid InputMethodEngineV4 instance."
+                     << "So it returns early.";
+        return;
     }
+
+    this4->virtualKeyboardEventImpl(entry, virtualKeyboardEvent);
+}
+
+FunctionMode InputMethodEngine::getFunctionMode() const {
```
wengxt, line 41: I don't quite understand why function mode needs a setter function. shouldn't it be a static property?

wengxt: It seems that the mode is a instance property, then shouldn't engine just read this property?

wengxt (changes requested):

src/lib/fcitx/inputcontext.h
```diff
@@ -41,6 +41,8 @@ class InputPanel;
 class StatusArea;
 typedef std::function<bool(InputContext *ic)> InputContextVisitor;
 
+enum class FunctionMode : uint32_t { Full = 1 << 0, Limited = 1 << 1 };
```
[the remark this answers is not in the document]
kingysu: OK. I will fix it.

src/lib/fcitx/inputcontext_p.h
kingysu: OK. I will fix it.

src/lib/fcitx/inputmethodengine.cpp
kingysu: The input method engine can support just full function mode or support both full function mode and limited funciton. So we should tell the input method engine which function mode should be effective. Or else, the input method engine should use full function mode when keyEvent function is caled and limited function mode when virtualKeyboardEvent function is called.

kingysu:
> Question, should function mode be a capability flag instead ? if capability flag bits are limited, we can introduce a new property for input context. Associate it to every key event seems wrong to me.

done

src/ui/virtualkeyboard/virtualkeyboard.cpp
```diff
@@ -120,6 +125,16 @@ class VirtualKeyboardBackend
     VirtualKeyboard *parent_;
 };
 
+void VirtualKeyboardBackend::setVirtualKeyboardFunctionMode(uint32_t mode) {
+    if (mode != static_cast<uint32_t>(VirtualKeyboardFunctionMode::Full) &&
+        mode != static_cast<uint32_t>(VirtualKeyboardFunctionMode::Limited)) {
+        return;
```
wengxt, line 131: throw invalid argument error here, 

See:
https://github.com/fcitx/fcitx5/blob/00d52a4a34aaabd2fac922deff1860c5ef04e759/src/modules/dbus/dbusmodule.cpp#L295

wengxt (changes requested):

src/ui/virtualkeyboard/virtualkeyboard.cpp
kingysu: fix it.

kingysu: > throw invalid argument error here,
> 
> See:
> 
> https://github.com/fcitx/fcitx5/blob/00d52a4a34aaabd2fac922deff1860c5ef04e759/src/modules/dbus/dbusmodule.cpp#L295

fix it.

PR #787, opened 2023-05-07
add immodule-probing for diagnose

xingwozhonghua:


data/fcitx5-diagnose.sh
```diff
@@ -250,6 +250,15 @@ print_process_info() {
     echo "$1 ${cmdline}"
 }
 
+check_software(){
+    local software=$1
+    if which $software >/dev/null 2>&1; then
```
wengxt, line 255: use command, instead of which

wengxt (changes requested):

PR #828, opened 2023-07-21
Add a new dbus method to get input method entries by group name

kingysu:
If the argument is an empty string, it will use the current group name, or else it will use the argument as the group name.

### DIFF
--- a/src/modules/dbus/dbusmodule.cpp
+++ b/src/modules/dbus/dbusmodule.cpp
@@ -182,6 +182,39 @@ public:
         return {"", {}};
     }
 
+    std::vector<
+        DBusStruct<std::string, std::string, std::string, std::string,
+                   std::string, std::string, std::string, bool, DBusVariantMap>>
+    getInputMethodEntriesByGroupName(const std::string &inputMethodGroupName) {
+        std::vector<DBusStruct<std::string, std::string, std::string,
+                               std::string, std::string, std::string,
+                               std::string, bool, DBusVariantMap>>
+            inputMethodEntries;
+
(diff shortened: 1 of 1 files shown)


src/modules/dbus/dbusmodule.cpp
```diff
@@ -271,6 +271,47 @@ class Controller1 : public ObjectVTable<Controller1> {
         return instance_->inputMethodManager().currentGroup().name();
     }
 
+    dbus::Variant
+    inputMethoEntry2Variant(const InputMethodEntry &inputMethodEntry) {
```
wengxt, line 275: the existing field are commonly used and are not optional field, so you should not put them into the variant map.

src/modules/dbus/dbusmodule.cpp
```diff
@@ -631,6 +672,8 @@ class Controller1 : public ObjectVTable<Controller1> {
                                "s", "");
     FCITX_OBJECT_VTABLE_METHOD(currentInputMethodGroup,
                                "CurrentInputMethodGroup", "", "s");
+    FCITX_OBJECT_VTABLE_METHOD(getIMInfoListByGroupName,
+                               "GetIMInfoListByGroupName", "s", "a{sv}");
```
wengxt, line 676: I expect the return value type to be a(sssssssba{sv})

which means it is an array of InputMethodEntry.

7 "s" & 1 "b" means the existing field, uniqueName etc.. (see AvailableInputMethods, that one is missing "addon" though)
a{sv} means the field reserved for the future, which can only be empty at this point.

wengxt (changes requested):

PR #854, opened 2023-08-28
fstat(3): Support large inode (>=2^32) on 32 bits system

sgn:


wengxt:
I don’t think this is the right approach. You should just set it with -D in cxxflags when compiling it.

PR #920, opened 2023-12-21
Some changes regarding macOS build

eagleoflqj:
* The linker options are not supported by the linker. Related: https://github.com/neovim/neovim/pull/23263
* Move `ENABLE_X11` to outer level as it makes no sense to generate `xim.conf`


wengxt:
@eagleoflqj test/ copy xim.conf unconditionally

Issue #938, opened 2024-01-17
tscancode静态代码检查


xingwozhonghua126:
<img width="1872" alt="Snipaste_2024-01-17_10-46-14" src="https://github.com/fcitx/fcitx5/assets/18661883/7ca70a8e-fc58-4a69-85e1-99133a40c361">

xingwozhonghua126:
![image](https://github.com/fcitx/fcitx5/assets/18661883/14b1bddb-3653-4ae6-86f5-6ba08120ee03)

PR #939, opened 2024-01-17
[dev] composeResult might be null

xingwozhonghua126:
https://github.com/fcitx/fcitx5/issues/938

### DIFF
--- a/src/im/keyboard/compose.cpp
+++ b/src/im/keyboard/compose.cpp
@@ -101,15 +101,18 @@ bool ComposeState::typeImpl(KeySym sym, std::string &result) {
     // check compose first.
     auto composeResult = instance_->processComposeString(inputContext_, sym);
     if (!composeResult) {
+        assert(composeBuffer_.size() > 0);
         if (instance_->isComposing(inputContext_)) {
             return true;
         }
         // If compose is invalid, and we are not in composing state before,
         // ignore this.
         if (!wasComposing) {
+            assert(composeBuffer_.size() == 1);
             composeBuffer_.clear();
             return false;
         }
+        assert(composeBuffer_.size() > 1);
 
         // Check for dead key prefix.
         size_t numDeadPrefix = 0;
@@ -155,6 +158,8 @@ bool ComposeState::typeImpl(KeySym sym, std::string &result) {
             reset();
             return true;
         }
+        // Unreachable.
+        return true;
     }
     // Empty string but composing, means ignored
     if (instance_->isComposing(inputContext_)) {
--- a/src/im/keyboard/compose.cpp
+++ b/src/im/keyboard/compose.cpp
@@ -165,11 +165,12 @@ bool ComposeState::typeImpl(KeySym sym, std::string &result) {
 
     reset();
     // Compose success or key should be ignored.
-    if (composeResult->empty()) {
+    if (composeResult && composeResult->empty()) {
         return false;
+    } else {
+        result.append(*composeResult);
     }
 
-    result.append(*composeResult);
     return true;
 }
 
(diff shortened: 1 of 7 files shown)


src/lib/fcitx/instance.cpp
```diff
@@ -1684,10 +1684,10 @@ std::string Instance::inputMethodIcon(InputContext *ic) {
     const auto *entry = inputMethodEntry(ic);
     auto *engine = inputMethodEngine(ic);
 
-    if (engine) {
+    if (engine && entry) {
```
wengxt, line 1687: This never happens, because valid engine ensures valid entry.

src/lib/fcitx/instance.cpp
```diff
@@ -1684,10 +1684,10 @@ std::string Instance::inputMethodIcon(InputContext *ic) {
...
         icon = engine->subModeIcon(*entry, *ic);
     }
-    if (icon.empty()) {
+    if (icon.empty() && entry) {
```
wengxt, line 1690: icon empty means subModeIcon returns null, which means there’s an valid entry.

src/ui/classic/xcbinputwindow.cpp
```diff
@@ -127,13 +127,12 @@ void XCBInputWindow::updateDPI(InputContext *inputContext) {
 }
 
 void XCBInputWindow::update(InputContext *inputContext) {
-    if (!wid_) {
+    if (!wid_ || !inputContext) {
```
wengxt, line 130: inputContext is allowed to be null and it will hide the window.

src/lib/fcitx/instance.cpp
xingwozhonghua126: done

src/lib/fcitx/instance.cpp
xingwozhonghua126: done

src/ui/classic/xcbinputwindow.cpp
xingwozhonghua126: done

src/im/keyboard/compose.cpp
```diff
@@ -170,11 +170,14 @@ bool ComposeState::typeImpl(KeySym sym, std::string &result) {
 
     reset();
     // Compose success or key should be ignored.
-    if (composeResult->empty()) {
```
wengxt, line 173: shouldn't require anymore. The all branch in the previous if check the null-ness.

src/modules/wayland/waylandmodule.cpp
```diff
@@ -314,11 +314,12 @@ bool WaylandModule::reopenConnectionSocket(const std::string &displayName,
 
     // Record all the IC in the old connection.
     std::vector<TrackableObjectReference<InputContext>> ics;
-    iter->second->focusGroup()->foreach([&ics](InputContext *ic) {
-        ics.push_back(ic->watch());
-        return true;
-    });
-
+    if (iter != conns_.end()) {
```
wengxt, line 317: fixed with an alternative way.

src/modules/xcb/xcbkeyboard.cpp
```diff
@@ -433,7 +433,7 @@ void XCBKeyboard::setRMLVOToServer(const std::string &rule,
     do {                                                                       \
         *data = FIELD##Len;                                                    \
         data += 1;                                                             \
-        if (FIELD##Len) {                                                      \
+        if (FIELD##Len && rnames.FIELD) {                                      \
```
wengxt, line 436: this is false positive since the null check is done above. Non-Zero length means not null.

PR #1004, opened 2024-04-02
add comment property to CandidateWord

eagleoflqj:
Currently comment of a rime candidate is [appended](https://github.com/fcitx/fcitx5-rime/blob/92a4410cfbcefd3782ec6b8efd54f9a071274f62/src/rimecandidate.cpp#L24-L27) into candidate in engine level, so it's hard to distinguish and apply different style.
The PR adds a new property to CandidateWord and preserves the behavior in 3 UI implementations (classicui and kimpanel tested, virtualkeyboard not tested).

### DIFF
--- a/src/lib/fcitx/candidatelist.h
+++ b/src/lib/fcitx/candidatelist.h
@@ -53,6 +53,14 @@ public:
      * @since 5.1.9
      */
     const Text &comment() const;
+    /**
+     * Return text with comment.
+     *
+     * @param separator separator between text and comment.
+     * @return value of comment.
+     * @since 5.1.9
+     */
+    Text textWithComment(std::string separator = " ") const;
 
 protected:
     void setText(Text text);
--- a/src/lib/fcitx/candidatelist.h
+++ b/src/lib/fcitx/candidatelist.h
@@ -46,12 +46,20 @@ public:
     bool isPlaceHolder() const;
     bool hasCustomLabel() const;
     const Text &customLabel() const;
+    /**
+     * Return comment corresponding to the candidate.
+     *
+     * @return value of comment.
+     * @since 5.1.9
+     */
+    const Text &comment() const;
 
 protected:
     void setText(Text text);
     void setPlaceHolder(bool placeHolder);
     void resetCustomLabel();
     void setCustomLabel(Text text);
+    void setComment(Text comment);
 
 private:
     std::unique_ptr<CandidateWordPrivate> d_ptr;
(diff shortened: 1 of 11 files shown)


wengxt:
updateClientSideUIImpl in dbusfrontend.cpp also need to be updated,

src/lib/fcitx/text.cpp
```diff
@@ -55,6 +55,14 @@ void Text::append(std::string str, TextFormatFlags flag) {
     d->texts_.emplace_back(std::move(str), flag);
 }
 
+void Text::append(const Text &text) {
```
wengxt, line 58: Can we make this function accept Text, so all the std::string could be moved from?

src/lib/fcitx/text.cpp
```diff
@@ -55,6 +55,14 @@ void Text::append(std::string str, TextFormatFlags flag) {
...
+    FCITX_D();
+    for (const auto &p : text.d_ptr->texts_) {
```
wengxt, line 60: This loop function can be replaced with std::copy  + std::make_move_iterator

src/lib/fcitx/text.h
```diff
@@ -37,6 +37,12 @@ class FCITXCORE_EXPORT Text {
     void setCursor(int pos = -1);
     void clear();
     void append(std::string str, TextFormatFlags flag = TextFormatFlag::NoFlag);
+    /**
+     * Append another text.
+     *
```
wengxt, line 42: missing `@param`

src/lib/fcitx/candidatelist.h
```diff
@@ -46,12 +46,20 @@ class FCITXCORE_EXPORT CandidateWord {
     bool isPlaceHolder() const;
     bool hasCustomLabel() const;
     const Text &customLabel() const;
+    /**
+     * Return comment corresponding to the candidate.
+     *
+     * @return value of comment.
+     * @since 5.1.9
+     */
+    const Text &comment() const;
```
wengxt, line 55: Can we also add a helper function, may be just called `Text textWithComment(std::string separator = " ") const` so we can save some duplicate code in UI.

wengxt (changes requested):

wengxt:
Also, whenever you add new API that was intended to be used by other project immediately, bump the version in CMakeLists.txt in the root if no one bumped the version after the previous release.

wengxt:
Please also add some tests that exercise the new api function in test/testtext.cpp and test/testcandidatelist.cpp

wengxt (approved):

PR #1070, opened 2024-06-05
add hotkey method to Action

eagleoflqj:
It will be nice to show hotkey bound to actions in macOS menu like built-in Pinyin does.

### DIFF
--- a/src/lib/fcitx/action.h
+++ b/src/lib/fcitx/action.h
@@ -9,6 +9,7 @@
 
 #include <memory>
 #include <fcitx-utils/element.h>
+#include <fcitx-utils/key.h>
 #include <fcitx-utils/macros.h>
 #include "fcitxcore_export.h"
 
@@ -138,6 +139,14 @@ public:
     void update(InputContext *ic);
 
     FCITX_DECLARE_SIGNAL(Action, Update, void(InputContext *));
+
+    /**
+     * Hotkey bound to the action.
+     *
+     * @return key list.
+     * @since 5.1.11
+     */
+    virtual KeyList hotkey() const { return {}; }
 
 private:
     void setName(const std::string &name);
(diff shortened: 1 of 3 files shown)


src/lib/fcitx/action.h
```diff
@@ -139,6 +140,14 @@ class FCITXCORE_EXPORT Action : public Element {
 
     FCITX_DECLARE_SIGNAL(Action, Update, void(InputContext *));
 
+    /**
```
wengxt, line 143: No, you can't add new virtual function without break ABI.

Let's just say we assume this is getter/setter type value and store it within ActionPrivate.

wengxt (changes requested):

src/lib/fcitx/action.h
wengxt: Also, in the documentation, mention this is for "Display only" if UI implementation supports it, and doesn't mean anything within the key handling logic.

src/lib/fcitx/action.h
```diff
@@ -139,6 +140,23 @@ class FCITXCORE_EXPORT Action : public Element {
...
+     * Hotkey bound to the action.
+     * This is only for display purpose when UI implementation supports it,
+     * and it has nothing to do with the key handling logic.
+     *
+     * @return key list.
+     * @since 5.1.11
+     */
+    KeyList hotkey() const;
```
wengxt, line 151: should be const reference

src/lib/fcitx/action.h
```diff
@@ -139,6 +140,23 @@ class FCITXCORE_EXPORT Action : public Element {
...
+
+    /**
+     * Set associated hotkey for display.
+     *
+     * @since 5.1.11
+     */
+    void setHotkey(const KeyList &);
```
wengxt, line 158: should have parameter name

wengxt (changes requested):

PR #1075, opened 2024-06-11
Ensure that the left of the input window is always visible

kingysu:
1. Make the code more clear
2. Ensure that the left of the input window is always visible


dontbethatguy209:
When performed used in cross platform does any compensation for left / right balance occur or does it mean I am right but I am left when you act a cross from me?

PR #1085, opened 2024-07-06
add API to indicate whether InputPanel is transient

eagleoflqj:
On macOS Terminal, a dummy client preedit is needed to close an empty clipboard by ESC or input ascii semicolon by Return with 双拼.
However, the dummy preedit shouldn't be set when using Shift to switch IM, as it clears current cell in LibreOffice.
The `setTransient` API could be used to distinguish these scenarios.


wengxt:
I don't like this design. So to me, there's some totally unrelated issue:
1. mac(or not just mac, I know firefox/ browser on all system having similar issues) need a dummy preedit to be able to workaround certain application issue. I believe the thing you want, is to make client be able to detect, whether input method is in Composing mode, even if client preedit is empty.

2. The usage of a OSD-like information popup with input panel which may be randomly modified.

The new API here solves none of them.

PR #1086, opened 2024-07-06
The virtual keyboard should show or hide automatically if enabled

kingysu:
1. When the user interacts with some applications rather than the virtual keyboard itself, the virtual keyboard should show or hide if it is enabled.

2. For example, if the user clicks a window and the window has a text input control which has the focus and accepts the input from the keyboard, then the virtual keyboard should show automatically if the auto show switch is enabled.

3. Then, if the window loses the focus, the virtual keyboard should hide automatically if the auto hide switch is enabled.

### DIFF
--- a/src/lib/fcitx/instance.cpp
+++ b/src/lib/fcitx/instance.cpp
@@ -1042,11 +1042,9 @@ Instance::Instance(int argc, char **argv) {
 
             activateInputMethod(icEvent);
 
-            if (virtualKeyboardAutoShow()) {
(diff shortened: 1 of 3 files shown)


wengxt:
I'm not fully understand the motivation of this change here.

To me, it seems like the "is auto hide" mode check is pushed down to the UI level. Which seems to be wrong to me.

UI's API should be simply pure, when show/hide is called, then it just show or hide, without checking any other "auto" condition.

As your issues, I think the issue should be solved in other some way.

PR #1087, opened 2024-07-06
The input method mode should be set physical keyboard when necessary

kingysu:
If the virtual keyboard should not show automatically, then the input method mode should be set physical keyboard when the user closes the virtual  keyboard by interacting with the virtual keyboard itself. For example, if the user clicks a close button in the virtual keyboard or clicks an icon in the task bar or a float button on the desktop which can close the virtual keyboard, then the virtual keyboard will be closed and the input method mode should be set physical keyboard.


wengxt:
I believe there's some missed expectation between your virtual keyboard and input method on what's the expected behavior. We probably need to chat offline to clarify all of these.

PR #1098, opened 2024-07-22
Allow users to cancel opening config directory

nuzar:
When fcitx configtool was not installed, fcitx will try to open it's configuration directory. Rather than edit config files, the user might want to install fcitx configtool properly then try again.


nuzar:
Test on kde
image

wengxt:
Thanks, I think it's a great idea.

PR #1124, opened 2024-08-26
Determine whether the obtained IC is the currently active IC, and then determine whether the engine needs to be inactivated

hantengc:
1. In most cases, focus out should be called first, and then focus in.
2. Before the input method engine deactivates, determine whether the obtained IC is the currently activated IC, and then decide whether to inactivate the engine. If it is not the currently activated IC, it does not need to deactivate the engine to avoid the inactivation of the input method engine

### DIFF
--- a/src/lib/fcitx/instance.cpp
+++ b/src/lib/fcitx/instance.cpp
@@ -2462,6 +2462,10 @@ void Instance::deactivateInputMethod(InputContextEvent &event) {
     if (!engine || !entry) {
         return;
     }
+
+    if (ic != mostRecentInputContext()) {
+        return;
+    }
     inputState->overrideDeactivateIM_ = entry->uniqueName();
     engine->deactivate(*entry, event);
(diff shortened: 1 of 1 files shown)


wengxt:
我已经解释过了，每个 focus group 都可以有 active ic

hantengc:
@wengxt 但是没有必要每次都通知引擎激活和失活吧？只有引擎失活的ic和激活的ic一致的时候，才认为引擎需要真正的失活

wengxt:
@hantengc deactivate是为了重置状态。事实上，你担心的部分一般都在界面那里处理

wengxt:
@hantengc 因为你错误的认为输入法引擎只有一个全局状态，实际上每个ic都应该保存
一个状态

wengxt:
@hantengc 参见 https://codedocs.xyz/fcitx/fcitx5/classfcitx_1_1InputContextProperty.html

PR #1132, opened 2024-09-12
fix Shift+Tab normalized to Tab

eagleoflqj:


wengxt:
please add a test case.

PR #1133, opened 2024-09-20
Optimize code

hantengc:
1. optimize icon acquisition logic to prepare for customizing icons for other desktop environments in the future

### DIFF
--- a/src/modules/notificationitem/notificationitem.cpp
+++ b/src/modules/notificationitem/notificationitem.cpp
@@ -53,20 +53,24 @@ public:
     }
     void activate(int, int) { parent_->instance()->toggle(); }
     void secondaryActivate(int, int) {}
-    std::string iconName() {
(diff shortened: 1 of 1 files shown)


CoelacanthusHex:
Shouldn't this thing be controlled by modifying the `input-keyboard` icon in the icon pack?

hantengc:
@CoelacanthusHex 1. Not sure if there are any other packages that use the input-keyboard icon, changing the system icon name will have other effects
2. In the past, I used to use icons for the KDE environment, but I think I could add icons for other desktops as well

wengxt:
I'm not sure if english us is the icon you want.

When this icon is used, it doesnt have to be English US

hantengc:
@wengxt It means that other layout input methods are also using this icon, right? Then I should have our UI designers design a more generic icon

PR #1145, opened 2024-10-05
support emscripten build

eagleoflqj:
Upstream patch from https://github.com/fcitx-contrib/fcitx5-js

### DIFF
--- a/src/lib/fcitx-utils/CMakeLists.txt
+++ b/src/lib/fcitx-utils/CMakeLists.txt
@@ -27,14 +27,16 @@ if (ENABLE_DBUS)
     endif()
 endif()
 
-if (NOT TARGET Systemd::Systemd)
+if (TARGET Systemd::Systemd)
+  set(FCITX_UTILS_SOURCES
+    ${FCITX_UTILS_SOURCES}
+    event_sdevent.cpp)
+elseif (TARGET ${LIBUV_TARGET})
   set(FCITX_UTILS_SOURCES
     ${FCITX_UTILS_SOURCES}
     event_libuv.cpp)
 else()
-  set(FCITX_UTILS_SOURCES
-    ${FCITX_UTILS_SOURCES}
-    event_sdevent.cpp)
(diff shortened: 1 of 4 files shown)


src/lib/fcitx-utils/CMakeLists.txt
```diff
@@ -27,14 +27,16 @@ if (ENABLE_DBUS)
...
 endif()
 
-if (NOT TARGET Systemd::Systemd)
+if (TARGET Systemd::Systemd)
   set(FCITX_UTILS_SOURCES
     ${FCITX_UTILS_SOURCES}
-    event_libuv.cpp)
+    event_sdevent.cpp)
+elseif (EMSCRIPTEN)
+  list(APPEND FCITX_UTILS_SOURCES event_js.cpp) # implemented in fcitx5-js
```
wengxt, line 35: I don't like having an external file dependency.

I think we could somehow, create a stub for EventLoop and having a global factory class to allow external event loop implementation.

eagleoflqj: I have no idea how to achieve that with a factory.
Added the event_js.cpp, since there is nothing specific to fcitx5-js. `emscripten_async_call` is generic API https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_async_call

wengxt: Without breaking the API, here's what you can do.

1. create a stub only EventLoop that takes a EventLoopImpl as parameter.
2. Create a global factory callback that allow external code to override the creation of EventLoopImpl.
3. Before initialize Instance, set the factory function

In other words, use Dependency injection to make your js eventloop implementation live outside of the main repo.

eagleoflqj: Updated.

eagleoflqj:
See https://github.com/fcitx-contrib/fcitx5-js/pull/14 for downstream build.

PR #1155, opened 2024-10-12
optimize code

hantengc:
…other desktop environments in the future

### DIFF
--- a/src/modules/notificationitem/notificationitem.cpp
+++ b/src/modules/notificationitem/notificationitem.cpp
@@ -53,20 +53,24 @@ public:
     }
     void activate(int, int) { parent_->instance()->toggle(); }
     void secondaryActivate(int, int) {}
-    std::string iconName() {
-        static bool preferSymbolic = !isKDE();
-        std::string icon;
-        if (preferSymbolic) {
-            icon = "input-keyboard-symbolic";
+    std::string keyboardIconName() const {
+        if (isKDE()) {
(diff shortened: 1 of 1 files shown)


hantengc:
@hantengc 这个pr

src/modules/notificationitem/notificationitem.cpp
```diff
@@ -53,20 +53,24 @@ class StatusNotifierItem : public dbus::ObjectVTable<StatusNotifierItem> {
     }
     void activate(int, int) { parent_->instance()->toggle(); }
     void secondaryActivate(int, int) {}
-    std::string iconName() {
-        static bool preferSymbolic = !isKDE();
-        std::string icon;
-        if (preferSymbolic) {
-            icon = "input-keyboard-symbolic";
+    std::string customIconForDesktop() const {
```
wengxt, line 56: since it's only about the keyboard icon, so just call it keyboardIconName or sth similar.

wengxt (changes requested):

src/modules/notificationitem/notificationitem.cpp
hantengc: > since it's only about the keyboard icon, so just call it keyboardIconName or sth similar.

好的，已修改~

hantengc:
修改改啦～

hantengc:
修改好了～

Issue #1162, opened 2024-10-25
XIM Preedit Area


xorgy:
### Summary

It is unclear to me whether the vanilla XIM protocol has any way for a client to report the preedit area to an IM server, but either way it would be nice if Fcitx5 had some way of accepting a preedit region (in addition to the caret position) via XIM.
I am interested in adding the required code to XIM clients (particularly winit and SDL3).

The documentation of X IM is very poor so I'm not entirely sure whether this already exists, or would need to be an extension.

xorgy:
I'd also be willing to write the patch for Fcitx5, but I want to make sure there's interest in accepting it before I put in the work.

wengxt:
Can you explain what's the meaning of it and how can it be used?

One thing that you may want to know is that, winit use libX11. You'll end up push changes to libx11.
I'm not sure about SDL, since they already dropped xim support in sdl 2.

Based on the protocol itself, if would be ok to add an new attribute. Attribute name is passed by string and im server will announce the supported attribute. See XIM_OPEN_REPLY in https://www.x.org/releases/X11R7.6/doc/libX11/specs/XIM/xim.html

xorgy:
其他的輸入法協議(iBus, Fcitx Dbus)有一個編輯區屬性有報告那個屬性的方法
Other IME protocols (iBus, Fcitx Dbus) have a way to report an area as the preedit area rectangle.

現在只有插入點的座標
Right now It is only the caret position: 

https://github.com/fcitx/fcitx5/blob/452672a0df6fa1ce4dee0361ffdec89e24066c8c/src/frontend/xim/xim.cpp#L325-L326

xorgy:
> Based on the protocol itself, if would be ok to add an new attribute. Attribute name is passed by string and im server will announce the supported attribute. See XIM_OPEN_REPLY in https://www.x.org/releases/X11R7.6/doc/libX11/specs/XIM/xim.html

Yeah, that may be the way to go, though there may be an issue with deciding when to report one or the other, and we would need to add state in the XIM frontend for Fcitx to ignore carets after receiving a rectangle.

wengxt:
I'm asking is because there's a XNArea in the protocol, that does something different (It's like negotiate with input method for a window size to render.)

https://docs.oracle.com/cd/E19620-01/805-3916/xtxlib-7/index.html

Yes, if you want cursor "point" to be cursor "rectangle", then it's indeed a new concept.

[...]

PR #1163, opened 2024-10-28
Allow XIM XNArea for client preedit area in OnTheSpot mode.

xorgy:


xorgy:
See #1162 for discussion and rationale.

xorgy:
@wengxt clang-format check is failing on files from master, you want a PR to fix this?

wengxt:
> @wengxt clang-format check is failing on files from master in CI, but runs fine locally. Not sure what to do with that.

CI uses clang 19 and there seems to be some regression on format. I tried to workaround it and now it should be fine.

PR #1169, opened 2024-11-04
notifications: add showTipV2 for extra context

eagleoflqj:
Intent is to show different notification type for rime deployment in fcitx5-ios and fcitx5-online.


eagleoflqj:
Using appIcon to indicate type instead.
https://github.com/fcitx-contrib/fcitx5-ios/commit/9c5d4119a4a86b362c81f6a5132320fadc4dae85

wengxt:
@eagleoflqj you should upstream those changes instead of maintaining lots of patches

PR #1174, opened 2024-11-07
make shared library loader optional

eagleoflqj:
iOS AppStore highly likely rejects app that calls dlopen with dynamic paths, by statically analyzing binary. So fcitx5-ios will link modules and engines as static libraries and remove references to dlopen.
Whoever wants to have dl on iOS without publishing is still free to do so.

## Summary by CodeRabbit

* **New Features**
  * Added a build option to enable or disable dynamic addon loading, enabled by default.
  * Builds configured without dynamic loading support no longer require the system’s dynamic-loading library.
  * Dynamic addon loading remains available when enabled, including library discovery, factory resolution, and error handling.

* **Bug Fixes**
  * Improved addon loading reliability by handling unavailable libraries and invalid addon factories gracefully.


eagleoflqj:
Not suitable to discard library.h which is used by fcitx5-lua, though with the same spirit to use dlopen

src/lib/fcitx/addonloader.cpp
```diff
@@ -14,7 +14,9 @@
 #include <utility>
 #include <vector>
 #include "fcitx-utils/flags.h"
+#ifndef FCITX_NO_DL
```
wengxt, line 17: Move this down to its own section, otherwise formatter won't sort the include before and after.

src/lib/fcitx/addonloader_p.h
```diff
@@ -13,7 +13,9 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#ifndef FCITX_NO_DL
```
wengxt, line 16: Might make it cleaner if you move shared library loader to its own file

PR #1182, opened 2024-11-16
make libuv optional for external eventloop impl

eagleoflqj:


wengxt:
No, I have some other plan for this

PR #1199, opened 2024-12-06
Add new macro and option to allow easy static build

wengxt:


eagleoflqj (approved):
Works on iOS with `*getStaticAddons()` as `REGISTRY_VARNAME` in `fcitx5_import_addons`.

PR #1220, opened 2024-12-30
Fix to prevent conflict between /etc/xdg/autostart/org.fcitx.Fcitx5.desktop and fcitx5 launched from KWin

phoepsilonix:
Added start-fcitx5.sh

Change org.fcitx.Fcitx5.desktop to a script that does not run fcitx5 when KWin's VirtualKeyboard is enabled.

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -4,6 +4,8 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/fcitx5-configtool.desktop.in.in
                ${CMAKE_CURRENT_BINARY_DIR}/fcitx5-configtool.desktop.in @ONLY)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/org.fcitx.Fcitx5.desktop.in.in
                ${CMAKE_CURRENT_BINARY_DIR}/org.fcitx.Fcitx5.desktop.in @ONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/start-fcitx5.sh.in
+               ${CMAKE_CURRENT_BINARY_DIR}/start-fcitx5.sh @ONLY)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/fcitx5-diagnose.sh
                ${CMAKE_CURRENT_BINARY_DIR}/fcitx5-diagnose ESCAPE_QUOTES @ONLY)
 
@@ -30,6 +32,9 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/fcitx5-diagnose" DESTINATION "${FCITX
 GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 
 install(DIRECTORY default DESTINATION "${FCITX_INSTALL_PKGDATADIR}" COMPONENT config)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/start-fcitx5.sh"
+        DESTINATION "${FCITX_INSTALL_LIBEXECDIR}/fcitx5/"
+        PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 
 if (WAYLAND_FOUND AND ENABLE_X11)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/20-fcitx-x11.in
--- a/data/org.fcitx.Fcitx5.desktop.in.in
+++ b/data/org.fcitx.Fcitx5.desktop.in.in
@@ -2,7 +2,7 @@
 Name=Fcitx 5
 GenericName=Input Method
 Comment=Start Input Method
-Exec=@FCITX_INSTALL_BINDIR@/fcitx5
+Exec=sh @FCITX_INSTALL_LIBEXECDIR@/fcitx5/start-fcitx5.sh
 Icon=@FCITX_ICON_NAME@
 Terminal=false
 Type=Application
--- a/data/start-fcitx5.sh.in
+++ b/data/start-fcitx5.sh.in
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [ "$XDG_CURRENT_DESKTOP" = "KDE" ];then
+if [ $(pgrep plasmashell) ];then
     SUDO_USER="$(who|head -n1|cut -f1 -d' ')"
     RET=$(su $SUDO_USER -c 'qdbus6 org.kde.KWin /VirtualKeyboard org.kde.kwin.VirtualKeyboard.active')
     if [ "$RET" = "true" ]; then
--- a/data/start-fcitx5.sh.in
+++ b/data/start-fcitx5.sh.in
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+if [ "$XDG_CURRENT_DESKTOP" = "KDE" ];then
+    SUDO_USER="$(who|head -n1|cut -f1 -d' ')"
+    RET=$(su $SUDO_USER -c 'qdbus6 org.kde.KWin /VirtualKeyboard org.kde.kwin.VirtualKeyboard.active')
+    if [ "$RET" = "true" ]; then
+        exit
+    fi
+fi
+exec @FCITX_INSTALL_BINDIR@/fcitx5


wengxt:
No, this is not a planned resolution.

Try use wayland launcher in virtual keyboard.

phoepsilonix:
Normally, even with X11, /etc/xdg/autostart/org.fcitx.Fcitx5.desktop is installed by `ninja install` with Wayland.

Then fcitx5 is launched. Is there any guarantee that fcitx5-wayland-launcher --reopen will be launched after that?

wengxt:
@phoepsilonix It doesn't launch any process by itself, it's dbus activation.

DBus will guarantee the one who win the dbus name will get the dbus call and establish the input method setup.

wengxt:
no it will still start. launcher will forward the socket via dbus. So, if xdg autostart starts earlier, it will get the dbus call and finish the setup.

If xdg starts later, dbus daemon will start fcitx process too. They will race but only one can win the org.fcitx.Fcitx5 name and receive the dbus call from wayland launcher. The one who loses will quit on its own.

phoepsilonix:
I see.
I will test again.
When installing fcitx5, I will only make changes to add the kwin configuration file, and leave autostart as is.
Then I will test it in the KDE Kwin environment.
I will check again to see if the Wayland diagnostics prompts me to start from Kwin.
Thanks for the information.

phoepsilonix:
@wengxt

Based on your advice, I modified the package so that the autostart file remains unchanged, and only the creation of the Kwin configuration file is done during the installation of fcitx5.

I understand that conflicts may arise, but there shouldn't be any issues.

During the installation of fcitx5, the following script adds the Kwin configuration.
```sh:fcitx5.install
setup_for_kde() {
    if [ $(pgrep plasmashell) ]; then
        echo "Setting up fcitx5 for KDE"
        [ -z $SUDO_USER ] && SUDO_USER=$(who|head -n1|cut -f1 -d" ")
        [ -z $SUDO_HOME ] && SUDO_HOME=$(eval echo ~$SUDO_USER)
        sudo -u $SUDO_USER mkdir -p $SUDO_HOME/.config/
        mkdir -p /etc/skel/.config/
        sudo -u $SUDO_USER kwriteconfig6 --file $SUDO_HOME/.config/kwinrc --group Wayland --type string --key InputMethod /usr/share/applications/fcitx5-wayland-launcher.desktop \
        || sudo -u $SUDO_USER kwriteconfig5 --file $SUDO_HOME/.config/kwinrc --group Wayland --type string --key InputMethod /usr/share/applications/fcitx5-wayland-launcher.desktop
        sudo -u $SUDO_USER sed -i 's/InputMethod/InputMethod[$e]/' $SUDO_HOME/.config/kwinrc
        if [ ! -e /etc/skel/.config/kwinrc ]; then
            touch /etc/skel/.config/kwinrc
        fi
        kwriteconfig6 --file /etc/skel/.config/kwinrc --group Wayland --type string --key InputMethod /usr/share/applications/fcitx5-wayland-launcher.desktop \
        || kwriteconfig5 --file /etc/skel/.config/kwinrc --group Wayland --type string --key InputMethod /usr/share/applications/fcitx5-wayland-launcher.desktop
        sed -i 's/InputMethod/InputMethod[$e]/' /etc/skel/.config/kwinrc
    fi
}

post_install() {
    setup_for_kde
    if [ $(pgrep plasmashell) ]; then
        echo
        echo 'Re-login or reboot to apply modify'
        echo
    fi
}

post_upgrade() {
    setup_for_kde
}

post_remove() {
    if [ $(pgrep plasmashell) ]; then
        [ -z $SUDO_USER ] && SUDO_USER=$(who|head -n1|cut -f1 -d" ")
        [ -z $SUDO_HOME ] && SUDO_HOME=$(eval echo ~$SUDO_USER)
        sudo -u $SUDO_USER kwriteconfig6 --delete --file $SUDO_HOME/.config/kwinrc --group 'Wayland' --key 'InputMethod' \
        || sudo -u $SUDO_USER kwriteconfig5 --delete --file $SUDO_HOME/.config/kwinrc --group 'Wayland' --key 'InputMethod'
        kwriteconfig6 --delete --file /etc/skel/.config/kwinrc --group 'Wayland' --key 'InputMethod' \
        || kwriteconfig5 --delete --file /etc/skel/.config/kwinrc --group 'Wayland' --key 'InputMethod'
        pgrep fcitx5 && pkill fcitx5 || true
    fi
}
```

Since installation scripts in Arch Linux's pacman, paru, yay, and Manjaro's pamac run in a sandbox environment, many environment variables cannot be retrieved. It also seems that obtaining information from DBUS and Wayland is not straightforward. Therefore, I determine that KDE is running by using `pgrep plasmashell`. Additionally, if SUDO_USER is unknown, I obtain the user performing the update using the `who` command.

If possible, would you consider adding fcitx5.install to the Arch fcitx5 package?

On a different note, I want to thank you for enabling the update of fcitx5-mozc after two years.
I appreciate your contribution and for submitting the MR. 

Thank you very much.

wengxt:
No, don't do that.

Arch's philosphy is to keep package vanilla.  Modify user level file is also a big NO.

If you really really want to do that, create your own AUR package. It can be empty in content, with just your post-install script.

wengxt:
@phoepsilonix I'm aware that you send your version of update to that, but to be honest it's not really well written and too complex and have too much customization. Such customization thing can be on AUR, but doesn't really belong to official repo.

I personally know the original maintainer but seems he's too busy with his life, so I decide to do my own version and keep it simple. The main issue I'd like to resolve is the icon name fallback in fcitx5-mozc.

phoepsilonix:
@wengxt
I personally haven't had any issues with the icons displaying, but maybe there's an issue with flatpak or something?

wengxt:
@phoepsilonix In latest version is fixed. flatpak is always fine. Refer to this commit https://github.com/fcitx/mozc/commit/e13338a41329f58e724801771f0cecf7338e24fb

Only when breeze icon theme is used because breeze has icon fcitx, but not fcitx-mozc due to xdg icon fallback, any fcitx-mozc* icon will be fallback to fcitx in breeze theme.

phoepsilonix:
@wengxt Are you saying that the breeze theme uses the fallback icon names by the theme and does not show the icons from the fcitx-mocz package?

wengxt:
This is what I mean

图片

This is what it should look like

图片

So basically how the icon name fallback works is that:
Try "a-b-c" in current theme, if not found, try "a-b" in current theme, if not found, try "a" in current theme, if not, try "a-b-c" in fallback theme, etc. https://specifications.freedesktop.org/icon-naming-spec/0.8.90/#guidelines

When an icon theme has fcitx, but not fcitx-mozc, the hicolor fcitx-mozc-* icon will be shadowed by the theme and only be dispalyed as "fcitx" icon.

phoepsilonix:
Manjaro's GNOME(Wayland,X11) default,dark style
xfce4 appearance theme(Matcha- etc)
KDE Plasma(Wayland,X11)'s breath, breath dark, breeze, breeze dark themes.

Switching between these themes and appearances did not cause any problems with icon display in my environment. The fcitx5-mozc icon is displayed in all themes.

wengxt:
@phoepsilonix latest fcitx5-mozc fixed that. if you want to reproduce you need breeze and 2.26

phoepsilonix:
So the problem is due to the version of breeze?
I can't reproduce it in the environment at hand, so if I encounter the same problem, I'll take some time out.

Have a good New Year.

wengxt:
@phoepsilonix no I mean fcitx5-mozc 2.26

phoepsilonix:
current Mozc is Mozc-2.30.5618.102+24.11.oss, do you mean the earlier 2.26 version?
That would require some sort of resignation, since they have stopped updating it.

For example, with Debian,Ubuntu, it's still an older version.
If they were to update, I think the assumption would be that, by policy, bazel or bazelisk would first be packaged in the official repositories, which is necessary for the build. If that is not possible, then they will continue to maintain the gyp build on their own.
They will also need to rethink the dependencies with submodule protobufs and other packages.
The assumption would be to include bazel or bazelisk in the official package and allow network access at build time, or to allow inclusion of external sources at build time that are not in the repository.

Unless we allow this, I doubt that new versions of Mozc will ever be included in the official repositories.
So, for now, the easier path is to provide new versions of Mozc in non-official repositories.

phoepsilonix:
I am not exactly sure of their policy. However, I think they need to include in their repository a set of sources that are necessary for the build. I am guessing that as a distributor, that is considered necessary to maintain the package.

Just so you know, the bazel build involves downloading multiple external sources. Including them in the official repositories is a procedural hassle and a complication since different versions have different dependencies.
Also, you may need to check submodule packages such as protobuf to make sure that they do not affect other packages besides Mozc.

For this reason, I am currently using the Arch system, which allows for easier packaging.
And for Debian and Ubuntu, I have built and published packages on github (one with an enhanced dictionary and one with a normal version).

I think it would be easier to access newer versions of Mozc if it were possible to download a set of external sources in a format similar to that of flatpak-fcitx5 and build them to provide them in external repositories, e.g. PPA?
However, I am not familiar with the procedures and steps involved, and it seems complicated, so I am not working on it.

hosiet:
Hi, Debian Developer and Debian's maintainer of fcitx5 ecosystem here. I can influence what is being provided via https://tracker.debian.org/pkg/mozc , but I did not follow this thread at all. Can I know what (as mozc upstream) do you expect us to do in the Debian/Ubuntu ecosystem?

Downloading from external website at build-time is not possible, of course. But we can always have dependency packaged in advance. It would be great if you could provide a list of such dependencies and we can work on them before upgrading mozc versions.

phoepsilonix:
@hosiet 
I'm glad that a Debian maintainer noticed this and commented on it.

One thing I'm concerned about is the compatibility of the submodules' protobuf, etc. In other words, can we expect that the git submodule source and the protobuf etc package version in the repository will always be the same?

It also seems that you are maintaining the gyp build, but the official source (upstream) is not maintaining it. Do you have any plans to officially provide feedback on the gyp build?

Are you considering migrating to the bazel build?
If you do migrate to the bazel build, please refer to mozc-deps.yaml in the https://github.com/fcitx/flatpak-fcitx5 repository. This will clarify the dependencies required for the build.

phoepsilonix:
In any case, the 2.26 version is probably bullseye in debian and 2.28 in bookworm. I don't think there are many places where the 2.26 version is used.

hosiet:
I went through the whole thread and it looks like the whole thread has shifted in its topic.

* The autostart issue has been thoroughly discussed.
* Regarding https://github.com/fcitx/fcitx5/pull/1220#issuecomment-2566177095 , I pushed the fix in Debian via cherry-picking fixes in the latest Debian-packaged versions (which is in Debian Testing/Unstable, or Debian 13). I will see if the fix can be applied on Debian Stable (Debian 12).
* Regarding putting bazel in Debian / Ubuntu: it is not possible in the foreseeable future without someone doing substantial work.
* Regarding protobuf: Still substantial work needed, as you can find out here: https://tracker.debian.org/pkg/protobuf .

Still, the policy of avoiding library vendoring and parallel-existing versions of library, and the recent drastic changes in protobuf, have blocked Debian / Ubuntu from obtaining a newer protobuf/mozc. The current version of mozc is stuck in 2023-08, before the new protobuf is required. I see that the bazel ecosystem tries to hardcode versions and handle everything, which conflicts in philosophy with distro-packaged versions. (CI build time / caching is meaningless in this scenario. We just want to have software built.) This conflict can only be mitigated and not completely solved.

phoepsilonix:
@hosiet
 Thanks for your comment.
I was able to confirm the situation.

I felt that if the submodule source was not split into packages but was treated as a single block of Mozc source, it might improve things a little. However, if that is difficult from a policy perspective, there is nothing that can be done.
Problems caused by differences in library version dependencies plague developers everywhere.

Well, this issue itself has already been closed.
Let's end the thread here.
I'm glad I was able to confirm the situation.

I hope this new year will be a good one for everyone. :smile:

PR #1235, opened 2025-01-27
define endian functions on win32

eagleoflqj:


wengxt:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort?view=msvc-170

PR #1236, opened 2025-01-27
define LC_MESSAGES when missing

eagleoflqj:
0 is LC_ALL on win32, although I don't think this else branch will be used in the future.


CoelacanthusHex:
It may be better to define LC_MESSAGES to LC_ALL for readability, since LC_ALL is a part of C standard.

PR #1242, opened 2025-01-27
disable geteuid on win32

eagleoflqj:
XDG_RUNTIME_DIR related code has no use case on win32.


wengxt:
I think you should plum this code to Windows/Temp or whatever user specific Temp.

PR #1243, opened 2025-01-28
implement getline on win32

eagleoflqj:


wengxt:
@eagleoflqj ....can you avoid extra copy of std::string buffer? that make the usage of getline pointless

wengxt:
I'd prefer avoid this change, not to mention it is poorly implemented.

I'll disable this API on win32 and replace all existing callsite with std::stream

PR #1248, opened 2025-02-01
Check inputContext before updating position

Johnson-zs:
Add null check for inputContext before calling updatePosition to prevent potential null pointer dereference in XCBInputWindow::update.


wengxt:
when it is null 

visible will be false and return in an earlier place.

Are you having a crash because of this?

Johnson-zs:
> when it is null
> 
> visible will be false and return in an earlier place.
> 
> Are you having a crash because of this?

Yes, it's really not necessary.

PR #1264, opened 2025-02-09
use semicolon as path delimiter on win32

eagleoflqj:


wengxt:
There is no plan to use any of these environment variables on windows.

PR #1273, opened 2025-02-19
Update CXX standard to 20

wengxt:
We will abandon ubuntu 20.04 and move to an new ubuntu. CI currently is on 22.04, though not sure if we want to jump to 24.04.

Some future plans on C++20 usage:
1. std::filesystem::path overloads in StandardPath
2. std::format in place of fmt
3. char8_t where appropriate
4. coroutine in some place for async data loading 
5. std::source_location
6. Heterogeneous lookup


yuyichao:
STD format is only available on gcc 13, which isn’t the default on Ubuntu 22.04 I believe…

wengxt:
> STD format is only available on gcc 13, which isn’t the default on Ubuntu 22.04 I believe…

then we can just jump to 24.04

PR #1284, opened 2025-02-26
fix: typo in clipboard.cpp password masking placeholder

wyu71:
- Fix typo "<Passowrd>" -> "<Password>" in clipboard.cpp
- Update corresponding translations in all .po files


wengxt:
only update code. pot will be updated with a cron job.

gostsdmitry:
#1285

PR #1306, opened 2025-03-24
clipboard: improve description of password display

pslldq:
Improve the description of the password display option. The previous english description was the wrong way around, as it indicated that the password was only shown when the checkbox was checked.


wengxt:
can you revert all the po change and keep only the code change.

it will be updated automatically.

pslldq:
adjusted it accordingly

PR #1333, opened 2025-04-21
Add Deepin desktop environment detection

wyu71:
Enhance the desktop environment detection logic to include support for the Deepin desktop environment.


wengxt:
1. Make sure this logic is what you really want: a string starts with "dde".
2. If so, use stringutils::startsWith

wyu71:
@wengxt stringutils::startsWith("dde") confirmed.

PR #1344, opened 2025-05-04
add keyboard shortcuts for changing to input method index within group

guoyiz23:
Tested on local machine on a `make install`, manually triggering `Alt+Shift+0`, `Alt+Shift+9`, `Alt+Shift+8`. Remains to add corresponding hotkey option to fcitx-config (something I do not feel especially excited about doing).

### DIFF
--- a/src/lib/fcitx/globalconfig.cpp
+++ b/src/lib/fcitx/globalconfig.cpp
@@ -143,6 +143,23 @@ FCITX_CONFIGURATION(
                                 _("Toggle embedded preedit"),
                                 {Key("Control+Alt+P")},
                                 KeyListConstrain()};
+    KeyListOption inputMethod0{this,
+                                "InputMethod0",
+                                _("Switch to input method 0 of input method group"),
+                                {Key("Alt+parenright")},
+                                KeyListConstrain()};
+    KeyListOption inputMethod1{this,
+                                "InputMethod1",
+                                _("Switch to input method 1 of input method group"),
+                                {Key("Alt+parenleft")},
+                                KeyListConstrain()};
+    KeyListOption inputMethod2{this,
+                                "InputMethod2",
+                                _("Switch to input method 1 of input method group"),
+                                {Key("Alt+asterisk")},
+                                KeyListConstrain()};
+    
+                                    
     Option<int, IntConstrain, DefaultMarshaller<int>, ToolTipAnnotation>
         modifierOnlyKeyTimeout{
             this,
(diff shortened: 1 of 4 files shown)


wengxt:
No.

input method selector addon can already do this.

guoyiz23:
> No.
> 
> input method selector addon can already do this.

fcitx5-config-screenshot

Thank you. Working now; however, in the config, the 2nd input method is also displayed as "1", though the hotkey works.

wengxt:
@guoyiz23 read your own code:

图片


Also, I'm not sure you understand what I mean, this input method selector, it provides two styles to do such things. One is same as yours and can even do more.

https://github.com/fcitx/fcitx5/tree/master/src/modules/imselector

图片

guoyiz23:
My bad. I had removed the following config options I added from the `ini` file.

```
[Hotkey/InputMethod0]
0=Alt+parenright

[Hotkey/InputMethod1]
0=Alt+parenleft

[Hotkey/InputMethod2]
0=Alt+asterisk
```

But those were still displayed in the config tool GUI. I had expected that one would need to do a separate update on the config tool, but now I would guess that instead via DBus, the config tool reads in all the options and loads corresponding GUI widgets.

```
~/.config/fcitx5 [1] 
$ which fcitx5-configtool 
/usr/bin/fcitx5-configtool

~/.config/fcitx5 [0] 
$ which fcitx5-config-qt
/usr/local/bin/fcitx5-config-qt
```

However, even with that option removed from `.ini` file, the buttons corresponding to that which I added are still there, not sure why, maybe after I uninstall the `fcitx5-config-qt` above, those buttons will no longer appear. Where are they "cached" exactly, if not `.config/fcitx5/config`? I had uninstalled my local build and upon running the one from apt, they still appear.

wengxt:
Did you restart fcitx itself? All config option are read from the running fcitx via dbus. There's no need to change configtool when there's a new option added.

Also check if you install it to another prefix like /usr/local .

guoyiz23:
> Did you restart fcitx itself? All config option are read from the running fcitx via dbus. There's no need to change configtool when there's a new option added.
> 
> Also check if you install it to another prefix like /usr/local .

Yes, I had restarted, specifically the one installed by apt but those buttons still appeared. 若愿意，也可以用中文交流。你这个 input method selector addon 的存在至少对我并不明显（当然，这个也可能是我软件资质便弱的问题），当时，基本只看了 global option。我input method selecton addon里添加了这些快捷之后，本来以为是那个 addon 自动将那些插入了 global options 的 GUI 里，没想到是原来我做的那些的遗留。

我用三个不同输入法，所以想要这个功能，Windows 的微软输入法明显显示将快捷绑定于特定输入法的配置功能。fcitx5 感觉挺复杂的，好多其功能我都不知道是啥。现在，我也想配置当改变输入法后不要在小框子里一秒钟显示 [en] 或 [五]，自己找早晚会找到，当然，您也一下子就能指出来。

过程中，还是熟悉了这代码仓及对应生态不少，多谢您的帮助。将来，我再试试为fcitx5做点更改也是有可能的。

wengxt:
有些非常用功能就是故意放在其他地方的。

PR #1359, opened 2025-06-07
fix shift+click changes im by checking MousePressed

eagleoflqj:
This should work better than the 250ms workaround.
MousePressed doesn't seem to be used elsewhere so could be a good fit for this purpose.
It's frontend's job to decide how this KeyState is set. On macOS, im doesn't receive mouse click event, but it can be determined by text selection change (although not all apps report it correctly, VSCode ✅ LibreOffice ❌).

### DIFF
--- a/src/lib/fcitx/instance.cpp
+++ b/src/lib/fcitx/instance.cpp
@@ -784,7 +784,9 @@ Instance::Instance(int argc, char **argv) {
             // Keep this value, and reset them in the state
             inputState->keyReleased_ = -1;
             const bool isModifier = origKey.isModifier();
-            if (keyEvent.isRelease()) {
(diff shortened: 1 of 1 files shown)


wengxt:
I feel you can use FcitxKey_VoidSymbol or FcitxKey_None for this.

eagleoflqj:
If you mean replacing the keysym, that would be hacky because fake information is passed. From fcitx's perspective, a shift is down and never up.

wengxt:
@eagleoflqj I mean, you send an extra fake no-op key that stops the logic. The logic assume there's no other key event between shift up and down.

eagleoflqj:
> send an extra fake no-op key that stops the logic

That makes sense.

PR #1390, opened 2025-07-21
[wayland] Do not show layout override warning when override is disabled

sophiel-meow:
This pull request consists of two commits:

1. Prevents the Wayland self-diagnose routine from issuing layout override warnings when "allow overriding system XKB settings" is disabled by the user. Previously, users would receive a confusing message about "sending keyboard layout configuration" even if layout override was explicitly disabled.

2. Refactors the layout diversity check logic in the same routine by replacing `std::unordered_set<std::string>` with `std::optional<std::string>`, which improves performance and clarity.

These changes improve both user experience and internal structure consistency, especially on non-KDE/GNOME compositors (e.g., Hyprland).

Tested with Fcitx5 under Arch Linux with Hyprland and verified the warning is correctly suppressed when override is disabled.

2025-07-21-220818_hyprshot

### DIFF
--- a/src/modules/wayland/waylandmodule.cpp
+++ b/src/modules/wayland/waylandmodule.cpp
@@ -718,23 +718,27 @@ void WaylandModule::selfDiagnose() {
 
         // layout diagnosis only when overriding is enabled
         if (*config_.allowOverrideXKB) {
-            std::unordered_set<std::string> groupLayouts;
+            std::optional<std::string> firstLayout;
             for (const auto &groupName :
                  instance_->inputMethodManager().groups()) {
                 if (const auto *group =
                         instance_->inputMethodManager().group(groupName)) {
-                    groupLayouts.insert(group->defaultLayout());
-                }
-                if (groupLayouts.size() >= 2) {
-                    messages.push_back(
-                        _("Sending keyboard layout configuration to wayland "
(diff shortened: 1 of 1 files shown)


wengxt:
Sorry I misunderstand the overrideXkb here, it's the wayland's config, not the global config one.

I think the change makes sense, but just FYI the notification service, that support "action', it should have a button says "don't show this again'.

It is actually supported by your notification service, but it can only be shown in a not very straight forward way. Check the manual of your notification.

src/modules/wayland/waylandmodule.cpp
```diff
@@ -716,20 +716,30 @@ void WaylandModule::selfDiagnose() {
...
-            if (groupLayouts.size() >= 2) {
-                messages.push_back(
-                    _("Sending keyboard layout configuration to wayland "
-                      "compositor from Fcitx is "
-                      "not yet supported on current desktop. You may still use "
-                      "Fcitx's internal layout conversion by adding layout as "
-                      "input method to the input method group."));
-                break;
+        // layout diagnosis only when overriding is enabled
+        if (*config_.allowOverrideXKB) {
```
wengxt, line 720: Can you just write it as if (!...) { return }

wengxt (changes requested):

PR #1391, opened 2025-07-23
Move keyboard grab reset after virtual keyboard reset when deactivating

xzn:
On hyprland, closing a virtual keyboard causes it to send release events for all keys that are still pressed. Currently fcitx5 close its vk after closing its hardware keyboard grab, meaning the release events end up being sent to the wrong place.

(while fcitx5 does try to release all the pressed key before closing its vk, the code is ?seemingly not triggered? when switching window focus. It's also after closing the keyboard grab so it wouldn't do the expected thing for this issue.)

As an example this fixes issue like [this one](https://github.com/H3rmt/hyprshell/issues/294)

Repo steps for a quick test:

1. run `wev`
2. focus on another window with text field active (fcitx5 running and active)
3. hold down a key then switch focus to `wev` while holding the key

Expected result (after this pr)
`wev` shows no key release event immediately.

Current result:
`wev` shows a key release event immediately even though the key is still pressed.

### DIFF
--- a/src/frontend/waylandim/waylandimserverv2.cpp
+++ b/src/frontend/waylandim/waylandimserverv2.cpp
@@ -170,7 +170,6 @@ WaylandIMInputContextV2::WaylandIMInputContextV2(
         ++serial_;
         if (pendingDeactivate_) {
             pendingDeactivate_ = false;
-            keyboardGrab_.reset();
             repeatInfo_.reset();
             // This is the only place we update wayland xkb mask, so it is ok to
             // reset it to 0. This breaks the caps lock or num lock. But we have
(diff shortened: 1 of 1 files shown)


xzn:
Hmm it's a bit subtle. I guess hyprland send those release events to avoid keys being stuck.

By diff do you mean hyprland should check the combined state of all keyboards and only send release event when all other keyboards are not pressed on that key?

On a related note, maybe hyprland need to update state and cleanup during keyboard grab destroy as well. Currently if persistent virtual keyboard is enabled in fcitx5, switching window focus between window with fcitx5 activated and window without, would lead to key getting stuck temporarily.

PR #1394, opened 2025-08-06
feat(dbus):add getAllHotkeys dbus interface

liehen:
The getAllHotkeys DBus interface returns information about global hotkeys and all hotkeys in the current input method group
The return value parameters include: category, name, unique name, description, and hotkey list

### DIFF
--- a/src/modules/dbus/dbusmodule.cpp
+++ b/src/modules/dbus/dbusmodule.cpp
@@ -28,6 +28,7 @@
 #include "fcitx-utils/dbus/variant.h"
 #include "fcitx-utils/eventloopinterface.h"
 #include "fcitx-utils/i18n.h"
+#include "fcitx-utils/key.h"
 #include "fcitx-utils/log.h"
 #include "fcitx-utils/misc.h"
 #include "fcitx-utils/stringutils.h"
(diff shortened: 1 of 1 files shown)


wengxt:
I'd rather this is implemented with multiple dbus call and filter on dbus client side.

wengxt:
If you insist on having a dbus call, you can implement in a separate (dbus) addon, instead of being an addition to this dbus addon.

Some example:
https://github.com/fcitx/fcitx5-rime/blob/master/src/rimeservice.cpp

The reason is I find this request is too specific and is tightly couple with whatever use case you have. It's useful, but not generally useful. If you find you have some other filtering request, you have to introduce something new. Instead, filter on cilent side is very flexible and you'll have much more freedom. I believe performance won't be a concern here, if it is indeed too slow, you can still do it with your own addon.

liehen:
Thank you for the suggestions. Below are the reasons for submitting this pull request：

Background: Using the UKUI-style input method configuration tool, there is currently no shortcut conflict detection. Implementing this functionality is required.

Current Problem: Although Fcitx5's DBus interface can retrieve the global configuration and individual input method configurations, the following issues exist:

After obtaining the configuration information, the contained shortcut keys need to be extracted.

Retrieving the input method configurations requires multiple calls to the GetConfig method.

To get the global shortcut configuration:
GetConfig("fcitx://config/global")

To get input method shortcut configurations: If there are 5 input methods in the current list, at least 7 interface calls are needed:
CurrentInputMethodGroup()
FullInputMethodGroupInfo("groupName")
GetConfig("fcitx://config/inputmethod/...") * 5 (once for each input method)

Therefore, a new getAllHotkeys method has been added to retrieve all shortcut information at once, enabling conflict detection and warnings through this data.

PR #1395, opened 2025-08-07
add loaded addons paths and  local data paths

hantengc:


wengxt:
StandardPath is deprecated, please check StandardPaths

PR #1398, opened 2025-08-12
start fcitx5 using a script if necessary

hantengc:
To make it better on different OS, you can start fcitx5 through a script.


wengxt:
I'd prefer you do such hack on downstream, where you have more flexibility to write a very customized script.

PR #1405, opened 2025-08-17
Add a new function to get locale

kingysu:


src/lib/fcitx/instance_p.h
```diff
@@ -133,6 +133,8 @@ class InstancePrivate : public QPtrHolder<Instance> {
     watchEvent(EventType type, EventWatcherPhase phase, EventHandler callback);
 
 #ifdef ENABLE_KEYBOARD
+    static std::optional<std::string> getLocale();
```
eagleoflqj, line 136: I think it's better to just return std::string, as the previous `std::optional` is because `getEnvironment` returns so. Of course, usages in `InstancePrivate::InstancePrivate` needs to be updated.

src/lib/fcitx/instance.cpp
```diff
@@ -224,6 +215,20 @@ InstancePrivate::watchEvent(EventType type, EventWatcherPhase phase,
 }
 
 #ifdef ENABLE_KEYBOARD
+std::optional<std::string> InstancePrivate::getLocale() {
+    const std::vector<const char *> variableList = {"LC_ALL", "LC_CTYPE",
```
eagleoflqj, line 219: Seems less overhead
`std::array variableList{"LC_ALL", "LC_CTYPE", "LANG"};`

src/lib/fcitx/instance.cpp
```diff
@@ -185,22 +185,13 @@ void InstanceArgument::printUsage() const {
 
 InstancePrivate::InstancePrivate(Instance *q) : QPtrHolder<Instance>(q) {
 #ifdef ENABLE_KEYBOARD
-    auto locale = getEnvironment("LC_ALL");
```
wengxt, line 188: Just use getCurrentLanguage in misc_p.h

wengxt (changes requested):

src/lib/fcitx/instance.cpp
wengxt: With some adaption to make it have the same behavior

PR #1413, opened 2025-08-30
Fix trash logfile path

iacore:
https://codeberg.org/iacore/fcitx5-weird-logfile


iacore:
Uh, it's fixed already?

PR #1415, opened 2025-09-07
Add default path

iacore:
I installed this library from AUR fcitx5-git. And it can't load any addon. Something must have been broken.

I think having a default value is better than having none.

### DIFF
--- a/src/lib/fcitx/addonloader.cpp
+++ b/src/lib/fcitx/addonloader.cpp
@@ -53,8 +53,9 @@ AddonInstance *SharedLibraryLoader::load(const AddonInfo &info,
             const auto libraryPaths = StandardPaths::global().locateAll(
                 StandardPathsType::Addon, file);
             if (libraryPaths.empty()) {
-                FCITX_ERROR() << "Could not locate library " << file
-                              << " for addon " << info.uniqueName() << ".";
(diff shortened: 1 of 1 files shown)


eagleoflqj:
The change here isn't the correct way to fix. You may start from https://github.com/fcitx/fcitx5/blob/13c36915d92c39657baf95b13dd24c422480728b/src/lib/fcitx-utils/standardpaths_p_unix.cpp#L120 to see how paths are configured and what default values are used.
Closing, but feel free to post your fcitx5-diagnose to see what's really going wrong.

wengxt:
I think there might be something wrong on how you build it. Can you describe your exact step of building the package?

wengxt:
There's no fcitx-git, only fcitx5-git.

I tried install it, and it works fine.

Did you ever set any extra environment variable on your system, specifically FCITX_ADDON_DIRS ?

iacore:
> There's no fcitx-git, only fcitx5-git.
> 
> I tried install it, and it works fine.
> 
> Did you ever set any extra environment variable on your system, specifically FCITX_ADDON_DIRS ?

Yes, `fcitx5-git`.

I did not set any extra environment variable. Is there any persistent configuration file used by fcitx5?

PR #1455, opened 2025-11-24
optimize the tray menu items

hantengc:
By referring to other OS, such as Windows and Mac, found that the settings function is always placed at the bottom. So did the same on fcitx5


wengxt:
doesn't seem to work as I expected on KDE. 

1. changing title it affects sorting in KDE tray, which is bad to me.
2. KDE's hover shows property Tooltip, instead of title.

PR #1456, opened 2025-11-26
replace delete this with safe async deletion in IBusInputContext

lzwind:
The previous code used 'delete this' in a service watcher callback, causing use-after-free when D-Bus clients disconnect unexpectedly.

Reproduction:
1. Start fcitx5 and IBus application (gedit)
2. Force kill: kill -9 $(pgrep gedit)
3. fcitx5 may crash due to use-after-free


wengxt:
the bug is on line 48, not really here

PR #1504, opened 2026-02-04
clipboard: allow searching the clipboard history

pslldq:
Implement a search functionality for the clipboard to allow searching in the clipboard history. Also don't impose a max history limit (of 30).

I've did some basic test and while the search is not optimized in any way, it works fine on ~21K entries without even maxing out a CPU core.

### DIFF
--- a/src/modules/clipboard/clipboard.cpp
+++ b/src/modules/clipboard/clipboard.cpp
@@ -18,6 +18,7 @@
 #include "fcitx-utils/event.h"
 #include "fcitx-utils/eventloopinterface.h"
 #include "fcitx-utils/i18n.h"
+#include "fcitx-utils/inputbuffer.h"
 #include "fcitx-utils/key.h"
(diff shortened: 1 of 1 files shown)


src/modules/clipboard/clipboard.cpp
```diff
@@ -256,16 +260,15 @@ Clipboard::Clipboard(Instance *instance)
 
             auto candidateList = inputContext->inputPanel().candidateList();
             if (candidateList) {
-                int idx = keyEvent.key().digitSelection();
+                int idx = keyEvent.key().digitSelection(KeyState::Alt);
```
wengxt: this breaks existing workflow. I'd prefer to just not allow search with numbers.

pslldq:
Hmm, this PR seems to be bugged a bit (for me?) as it still shows the old diff despite the branch being updated: https://github.com/fcitx/fcitx5/compare/master...pslldq:fcitx5:betterclip

PR #1576, opened 2026-06-08
Allow Enter key to commit selected candidate in quickphrase

### LINKED ISSUE #1443: commit selection using Enter/Return key in Quickphrase
### Summary

hi, 
it looks like the only way to commit a selected option in quickphrase is by pressing Spacebar. Pressing Return/Enter exits the quickphrase input method. I have unsuccessfully tried to adapt myself but months later I still find myself pressing Enter when I meant to commit.
Using Enter/Return to commit works fine in Unicode input mode. 

request: setting Enter/Return as a commit key along with Spacebar, or an option to use the Enter key to commit a suggestion from quickphrase.

wyu71:
## Summary
- Allow Enter/Return key to commit the highlighted candidate in quickphrase, matching the existing Space key behavior
- Original Enter behavior (commit raw input buffer) remains as fallback when no candidates are present
- Extract `commitCandidate` lambda to avoid duplication between Space and Enter handling

Fixes #1443

### DIFF
--- a/src/modules/quickphrase/quickphrase.cpp
+++ b/src/modules/quickphrase/quickphrase.cpp
@@ -137,13 +137,24 @@ QuickPhrase::QuickPhrase(Instance *instance)
                     }
                 }
 
-                if (keyEvent.key().check(FcitxKey_space) &&
-                    !candidateList->empty()) {
+                auto commitCandidate = [&]() {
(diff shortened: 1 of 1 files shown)


wyu71:
Closing, will resubmit.

PR #1578, opened 2026-06-11
fix(notificationitem): show current input method name in SNI tooltip

wyu71:
## Summary
- Fix `notifyNewTooltip()` to compare actual tooltip content instead of the constant `title()` return value
- `notifyNewTooltip()` was comparing against `title()` which always returns `"Input Method"`, so `NewToolTip` D-Bus signal was never emitted and the tooltip never updated on input method switch

### DIFF
--- a/src/modules/notificationitem/notificationitem.cpp
+++ b/src/modules/notificationitem/notificationitem.cpp
@@ -112,22 +112,20 @@ public:
 
     std::string label() { return ""; }
 
-    std::string title() { return _("Input Method"); }
+    std::string title() {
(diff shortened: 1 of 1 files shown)


wengxt:
图片

Title is shown here under, which seems to be better to be constant string.

wyu71:
@wengxt ，Thanks for the feedback. Updated: `title()` and `tooltip()` are reverted to original. The only change now is in `notifyNewTooltip()` — it was comparing against `title()` which always returns the constant "Input Method", so the `NewToolTip` signal was never emitted. Now it compares the actual tooltip content (`tooltip()`'s title field) instead.

PR #1583, opened 2026-06-17
waylandim: expose raw zwp_input_method_v2 to external UI addons

titoo-dev:
## Summary

Expose the raw `zwp_input_method_v2` of the current input context to external
UI addons, and install the `WaylandIM` public module so they can link it.

This lets a third-party UI addon create a compositor-positioned
`zwp_input_popup_surface_v2` (caret tracking) on Wayland — currently impossible
outside the `waylandim` frontend, because the only accessor returns the internal
`fcitx::wayland::ZwpInputMethodV2` wrapper whose header is not installed.

## Changes

- Add `zwp_input_method_v2 *WaylandIMModule::getInputMethodV2Raw(InputContext *)`
  — a one-liner: `rawPointer(getInputMethodV2(ic))`.
- Export it as a published addon function.
- `INSTALL` the already-existing `WaylandIM` exported CMake module +
  `waylandim_public.h` (currently build-tree only).

## Compatibility

Purely additive: no signature/behaviour changes, no new dependencies. Existing
`getInputMethodV2` consumers are unaffected. This mirrors how the `wayland`
module already installs its `wayland_public.h`, making the input-method frontend
extensible the same way.

## Motivation

Built a predictive input method (engine + n-gram daemon + Qt Quick candidate
bar). The candidate bar needs caret positioning via `get_input_popup_surface`;
without this, it requires patching/forking fcitx5 per distro. This change lets
such UI addons ship as plain addons on any distro.

### DIFF
--- a/src/frontend/waylandim/CMakeLists.txt
+++ b/src/frontend/waylandim/CMakeLists.txt
@@ -27,4 +27,4 @@ fcitx5_translate_desktop_file(${CMAKE_CURRENT_BINARY_DIR}/waylandim.conf.in wayl
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/waylandim.conf" DESTINATION "${FCITX_INSTALL_PKGDATADIR}/addon"
         COMPONENT config)
 
-fcitx5_export_module(WaylandIM TARGET waylandim BUILD_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}" HEADERS waylandim_public.h)
+fcitx5_export_module(WaylandIM TARGET waylandim BUILD_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}" HEADERS waylandim_public.h INSTALL)
--- a/src/frontend/waylandim/waylandim.cpp
+++ b/src/frontend/waylandim/waylandim.cpp
@@ -70,6 +70,10 @@ wayland::ZwpInputMethodV2 *WaylandIMModule::getInputMethodV2(InputContext *ic) {
         ->inputMethodV2();
 }
 
+zwp_input_method_v2 *WaylandIMModule::getInputMethodV2Raw(InputContext *ic) {
+    return wayland::rawPointer(getInputMethodV2(ic));
+}
+
 bool WaylandIMModule::hasKeyboardGrab(const std::string &display) const {
     if (const auto *server = findValue(servers_, display); server && *server) {
         if (server->get()->hasKeyboardGrab()) {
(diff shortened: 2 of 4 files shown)


CoelacanthusHex:
I don't think it's right way. If UI addons will need some information or API which are not exposed by fcitx5, you should expose the concrete one instead of exposing the whole internal zwp_input_method_v2 object. And all you need (zwp_input_popup_surface_v2) seems has been provided by `getInputPopupSurface` API and Classic UI addon has used it.

wengxt:
getInputMethodV2 already gives you the relevant pointer. You'd make a copy of fcitx-wayland to match the signature and use it.

But what you'd pay attention to is that under fcitx arch the ui addon is shared among every user. If you have any specialized use case, you'd consider do sth different.

Also I took a peek on your code. You seem to want to use Qt for UI.  The way that you're using it would be break for many things (e.g. timer). 

You'd consider create your own main function to launch fcitx instead. Create QApplication there, can make use of EventLoop::setEventLoopFactory to use Qt's event loop to drive fcitx's event loop. That way you can make both fcitx and Qt event works.

hantengc:
The issue this PR addresses probably bothers a lot of developers. It could be due to the Wayland protocol or some other reason, but right now, under Wayland, there are always problems with displaying the input method candidate window. The best solution is to leave it to the compositor, but the protocol can only have one registrant.

CoelacanthusHex:
> The issue this PR addresses probably bothers a lot of developers. It could be due to the Wayland protocol or some other reason, but right now, under Wayland, there are always problems with displaying the input method candidate window. The best solution is to leave it to the compositor, but the protocol can only have one registrant.

Why don’t you read the code at all? There has been a API `ZwpInputMethodV2::getInputPopupSurface()` to get `ZwpInputPopupSurfaceV2` which is a thin wrapper of `zwp_input_popup_surface_v2`.
https://github.com/fcitx/fcitx5/blob/7d5aa1b7be1fe40e3b83ad3b641d8f3d61522aef/src/lib/fcitx-wayland/input-method-v2/zwp_input_method_v2.h#L34
Classic UI addon also give you an example to use it.
https://github.com/fcitx/fcitx5/blob/7d5aa1b7be1fe40e3b83ad3b641d8f3d61522aef/src/ui/classic/waylandinputwindow.cpp#L194
I have mentioned this API in https://github.com/fcitx/fcitx5/pull/1583#issuecomment-4729511165. Why you never try to search `getInputPopupSurface` string in code? Its name has informed its purpose, and it only have simple three search result, one declaration, one implementation, and one usage. Why you never check it and just say "No way to do currently"? Why?

PR #1596, opened 2026-06-25
fix: notify compositor when preedit becomes empty in waylandim v2

SHORiN-KiWATA:
### Environment
- OS: Arch Linux, Linux 7.0
- Display: Wayland + niri
- fcitx5: 5.1.20
- Affected apps: Electron apps using wayland_v2 IM protocol (linuxqq, VS Code, etc.)

### Reproduction
1. Use an Electron app with Wayland IM protocol (e.g., linuxqq with `--ozone-platform=wayland`)
2. Type several characters (preedit appears)
3. Press Backspace repeatedly to delete
4. When only one character remains, press Backspace: the candidate window disappears, but the character remains displayed
5. Press Backspace once more: the character is finally deleted
2026-06-25-200012-REGION

### Root Cause
`updatePreeditDelegate` only calls `setPreeditString()` when `preedit.textLength() > 0`. When the last character is deleted, `textLength()` becomes 0, so the update is skipped. The Wayland compositor is never notified that the preedit is now empty, leaving the old preedit display on screen. A second Backspace is needed to trigger a state change that eventually clears it.

### Fix
Change `> 0` to `>= 0`. An empty preedit string is sent to the compositor, which correctly clears the display.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input composition handling so empty preedit text is now processed correctly, helping avoid missed updates in text entry.
  * Kept the existing safety limit in place for longer preedit content.

### DIFF
--- a/src/frontend/waylandim/waylandimserverv2.cpp
+++ b/src/frontend/waylandim/waylandimserverv2.cpp
@@ -685,7 +685,7 @@ void WaylandIMInputContextV2::updatePreeditDelegate(InputContext *ic) const {
     }
 
     // Validate not empty and within wayland limit.
-    if (preedit.textLength() > 0 &&
+    if (preedit.textLength() >= 0 &&
         preedit.textLength() < WaylandIMServerBase::safeStringLimit) {
         if (cursorStart < 0) {
             cursorStart = cursorEnd = preedit.textLength();


wengxt:
This is really compositor problem or application problem. Since 

1. empty preedit, is not really null preedit. this can be problematic on application side to think that input method is not done. Your change it not acceptable as it will break other application.
2. application doesn't talk in zwp_input_method_v2, input method do. compositor converts input method implementation to certain text-input-v1,2,3.

If you take a look on text_input_v3, it has allow null, 

```
<event name="preedit_string"> 
  <arg name="text" type="string" allow-null="true"/>
  <arg name="cursor_begin" type="int"/>
  <arg name="cursor_end" type="int"/>
</event>
```

but instead, on zwp_input_method_v2

```
<event name="set_preedit_string"> 
  <arg name="text" type="string" />
  <arg name="cursor_begin" type="int"/>
  <arg name="cursor_end" type="int"/>
</event>
```

Similar, but missing the key part "allow null". The only way to do this in input method v2 is to not send that instead of send empty.

It is compositor's job to convert it to the corresponding null preedit message for null.

Also it works pretty normal under kwin.

This is what kwin does

```
[1426666.055] zwp_text_input_v3#44.preedit_string("f", 1, 1)
[1426666.069] zwp_text_input_v3#44.done(7)
[1427717.069] zwp_text_input_v3#44.preedit_string(nil, 0, 0)
[1427717.076] zwp_text_input_v3#44.done(7)
````

PR #1598, opened 2026-06-28
fix: anchor candidate panel at cursor bottom to prevent covering input fields on GNOME Shell/Wayland

TerryLiu:
### Problem

On GNOME Shell/Wayland, the candidate panel rendered by the kimpanel
D-Bus protocol covers the input field instead of appearing below it.
This affects all applications that use `zwp_text_input_v3` (e.g.,
Chrome, GTK 4 apps).
截图 2026-06-28 19-13-29

### Maybe Cause

In `KimpanelProxy::updateCursor()`, the `SetSpotRect` D-Bus message
sent to GNOME Shell uses `cursorRect().top()` as the Y coordinate.
GNOME Shell's kimpanel implementation anchors the candidate panel at
the given (x, y) position, so using the cursor's top edge causes the
panel to overlap the input field.

### Fix

**`src/ui/kimpanel/kimpanel.cpp`** — `KimpanelProxy::updateCursor()`:

1. **Y coordinate**: Changed from `cursorRect().top()` to
   `cursorRect().bottom() + 80px`.

2. **Minimum size**: Enforce `w >= 1, h >= 1` because Chrome via
   `zwp_text_input_v3` can send cursor rectangles with 0px width,
   which interferes with GNOME Shell's positioning heuristics.

**Before:**
```cpp
int32_t y = inputContext->cursorRect().top();
```

**After:**
```cpp
static constexpr int32_t kCursorMargin = 80;
int32_t y = inputContext->cursorRect().bottom() + kCursorMargin;
```

### Data Flow

```
Chrome (zwp_text_input_v3) → GNOME Shell → IBus D-Bus
  → fcitx5 ibus frontend → setCursorRect()
  → kimpanel → SetSpotRect D-Bus → GNOME Shell renders panel
```

fcitx5 does not directly position the panel on GNOME Wayland — it
sends coordinates to GNOME Shell via `SetSpotRect`, and GNOME Shell
decides the final placement.

### Why 80px?

The offset was determined through gradient testing across multiple
scenarios. Key observations during testing:

1. **GNOME Shell's SetSpotRect response is non-linear.** Sending +200
   only moves the panel ~40-60px on screen. GNOME Shell applies
   internal clamping to the coordinates.

2. **Spot rect dimensions affect GNOME Shell's positioning.** Reducing
   the spot rect to 1×1 pixel caused the panel to move *upward* rather
   than down, indicating that GNOME Shell uses the rect's height in
   its positioning heuristic.

3. **Chrome's bookmarks bar shifts web content.** When the bookmarks
   bar is visible, page Y coordinates shift but GNOME Shell's cursor
   position calculation does not fully compensate, causing the panel
   to appear higher relative to the input field. The 80px offset
   covers both cases (with/without bookmarks bar).

| Offset | Chrome omnibox | Web inputs (no bookmarks) | Web inputs (with bookmarks) |
|--------|---------------|---------------------------|-----------------------------|
| original (top) | covered | covered | covered |
| bottom()+0 | slightly better | covered | covered |
| bottom()+8 | improved | covered | covered |
| bottom()+16 | — | — | moved *up* (worse) |
| bottom()+50 | — | correct | lower half covered |
| bottom()+200 | too far | too far | — |
| **bottom()+80** | ✅ | ✅ | ✅ |

### Additional: Diagnostic Logging

Added `FCITX_INFO` / `CLASSICUI_INFO` logging at each layer of the
positioning pipeline to aid future debugging:

| File | Location |
|------|----------|
| `src/lib/fcitx/inputcontext.cpp` | `setCursorRect()` |
| `src/ui/kimpanel/kimpanel.cpp` | `updateCursor()`, `update()` |
| `src/ui/classic/classicui.cpp` | `update()` |
| `src/ui/classic/waylandinputwindow.cpp` | `update()` |
| `src/ui/classic/inputwindow.cpp` | `update()` |
| `src/ui/classic/xcbinputwindow.cpp` | position calculation |

### Build Fix

Relaxed `wayland-protocols` required version from 1.46 to 1.45 for
compatibility with Ubuntu 24.04.

### Limitations

This is a **workaround**, not a root-cause fix.

The root cause lies in **GNOME Shell's kimpanel implementation**: it
places the candidate panel at the exact (x, y) coordinates received
via `SetSpotRect`, without checking whether the panel will cover the
input field or overflow the screen bounds.

A proper fix would require GNOME Shell to:
1. Render the panel **below** the cursor, not at the cursor's top
2. Perform **collision detection** — check if there is enough space
   below; if not, flip the panel above the cursor
3. Respect screen boundaries and clamp appropriately

The 80px offset compensates for this on a specific setup (GNOME Shell
default theme on Ubuntu 24.04, 1920×1080 + 2048×1280 monitors). It may
not be optimal for users with different themes, font sizes, or screen
layouts, because fcitx5 has no way to query GNOME Shell's actual
rendered panel height.

### Future Work

- Make the vertical offset configurable via `kimpanel.conf` (e.g.,
  `VerticalOffset` option), so users can tune it for their setup
  without recompiling.
- File an issue upstream against GNOME Shell / Mutter proposing that
  the kimpanel implementation perform cursor-relative positioning with
  automatic collision detection.

### Verified On

- **OS**: Ubuntu 24.04
- **DE**: GNOME Shell (Wayland)
- **Monitors**: Two vertically stacked screens (HDMI-1 1920×1080 above
  eDP-2 2048×1280)
- **Tested scenarios**:
  - Chrome navigation bar (omnibox)
  - Google search input at top of page
  - Web page input fields at various positions
  - Chrome with and without bookmarks bar

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ if (ENABLE_WAYLAND)
     set(REQUIRED_WAYLAND_COMPONENTS Client Cursor)
     find_package(Wayland 1.22 REQUIRED COMPONENTS ${REQUIRED_WAYLAND_COMPONENTS})
     find_package(WaylandScanner REQUIRED)
-    find_package(WaylandProtocols 1.46 REQUIRED)
+    find_package(WaylandProtocols 1.45 REQUIRED)
     pkg_check_modules(Gio IMPORTED_TARGET gio-2.0)
 
     if (USE_SYSTEM_PLASMA_WAYLAND_PROTOCOLS)
--- a/src/lib/fcitx/inputcontext.cpp
+++ b/src/lib/fcitx/inputcontext.cpp
@@ -268,6 +268,12 @@ void InputContext::setCursorRect(Rect rect, double scale) {
     if (d->cursorRect_ == rect && d->scale_ == scale) {
         return;
     }
+    FCITX_INFO() << "[CursorRect] program=" << program()
+                 << " frontend=" << frontendName()
+                 << " display=" << display()
+                 << " oldRect=" << d->cursorRect_
+                 << " newRect=" << rect
+                 << " scale=" << scale;
     d->cursorRect_ = rect;
     d->scale_ = scale;
     d->emplaceEvent<CursorRectChangedEvent>(this);
--- a/src/ui/classic/classicui.cpp
+++ b/src/ui/classic/classicui.cpp
@@ -456,10 +456,11 @@ void ClassicUI::update(UserInterfaceComponent component,
             ui = uiPtr->get();
         }
     }
-    CLASSICUI_DEBUG() << "Update component: " << static_cast<int>(component)
+    CLASSICUI_INFO() << "Update component: " << static_cast<int>(component)
                       << " for IC program:" << inputContext->program()
                       << " frontend:" << inputContext->frontendName()
                       << " display:" << inputContext->display()
+                      << " cursorRect:" << inputContext->cursorRect()
                       << " ui:" << (ui ? ui->name() : "(not available)");
 
     if (ui) {
--- a/src/ui/classic/inputwindow.cpp
+++ b/src/ui/classic/inputwindow.cpp
@@ -457,6 +457,11 @@ std::pair<int, int> InputWindow::update(InputContext *inputContext) {
             visible_ = false;
         }
     }
+    CLASSICUI_INFO() << "[InputWindow] nCandidates=" << nCandidates_
+                     << " upperLen=" << pango_layout_get_character_count(upperLayout_.get())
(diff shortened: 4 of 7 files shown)


wengxt:
1. it works perfectly fine here, especially for chrome / gtk4 ones you mentioned

image

I don't think your fix makes sense. If there's anything need to be adjusted, it should be in kimpanel itself. https://github.com/wengxt/gnome-shell-extension-kimpanel/

Also, there's some mutter regression recently:
https://gitlab.gnome.org/GNOME/mutter/-/work_items/4854
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5122

2. your code is with too much nonsense change. indicates you don't even read after whatever AI generate it for you. Please do your own house cleaning before send PR.

TerryLiu:
You're right.

Issue #845, opened 2023-08-15
Hotkeys bubble up to an application


andrhua:
**Describe the bug**
I want "Enumerate Input Method Forward" to be right alt key, and it works given the currently focused application doesn't recognize Alt as shorcut. Unfortunately, some graphical apps like Firefox and Slack bring up menu on alt press, and it's not always possible to disable this behavior.

**To Reproduce**
1. Configure IM switch to be alt
2. Open firefox/slack
3. Press alt

**Expected behavior**
Hotkeys, at least IM switching ones, do not bubble up to other apps, similarly to how `grp:..` xorg options work - once you set caps lock to switch layout, you basically lose the ordinary caps lock functionality.

**Desktop (please complete the following information):**
 - Desktop: i3
 - Display server type: X11

wengxt:
such thing is unavoidable, since under X11, usually it's application has higher priority to handle the key.

I'd suggest you try right ctrl or use a combination e.g. alt + shift.

wengxt:
An alternative solution is to use group switching in fcitx, which does use key grab. But I found some bug that prevent this option to work with modifier keys, which was fixed in 191660bb.

If you want to try that, simply create multiple groups and put only 1 input method within each group.

PR #1616, opened 2026-07-26
Add option to consume redundant Activate/Deactivate key press

akovalenko:
The current behavior — an Activate/Deactivate key press that doesn't
change state falls through to the application — is clearly intentional
and fine for most setups (as discussed in #845), so this PR changes no
defaults.

My use case is a modeless switch: dedicated `ActivateKeys=Super+N` /
`DeactivateKeys=Super+S`, picked to be idempotent — a redundant press
should be a no-op. Instead it leaks into the client and types a letter:
with the IM already active, Super+N inserts "n" (easy to reproduce with
two input methods in a group).

This adds an opt-in Hotkey option "Consume redundant Activate/Deactivate
key" (`ConsumeRedundantActivateKeys`, default false). When enabled, a
non-modifier press of an activate/deactivate key that no state-gated
handler claimed is consumed after the handler loop instead of being
forwarded. Modifier-only keys are never consumed (a bare modifier press
still reaches the client), and a key present in both lists (toggle
setup) matches a handler earlier and is unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added an option to consume redundant Activate/Deactivate key events when the input method is already in the requested state.
  * The setting is disabled by default and does not affect modifier-only keys.

* **Bug Fixes**
  * Prevented redundant activation or deactivation keystrokes from being passed to applications when the option is enabled.

### DIFF
--- a/src/lib/fcitx/globalconfig.cpp
+++ b/src/lib/fcitx/globalconfig.cpp
@@ -72,6 +72,19 @@ FCITX_CONFIGURATION(
         {Key("Hangul_Romaja")},
         KeyListConstrain({KeyConstrainFlag::AllowModifierLess,
                           KeyConstrainFlag::AllowModifierOnly})};
+    OptionWithAnnotation<bool, ToolTipAnnotation>
+        consumeRedundantActivateKeys{{
+            .parent = this,
+            .path{"ConsumeRedundantActivateKeys"},
+            .description{_("Consume redundant Activate/Deactivate key")},
+            .defaultValue = false,
+            .annotation{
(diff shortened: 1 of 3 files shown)


wengxt:
I think there the code can be much simpler: 

change canActivate / canDeactivate's return value to be inputState->isActive() || your new option.

akovalenko:
> I think there the code can be much simpler:
> 
> change canActivate / canDeactivate's return value to be inputState->isActive() || your new option.

Thank you, tried your approach, much simpler indeed. Works for me for several days. PR updated.

PR #1620, opened 2026-07-31
Add HiDPI support to classic UI XCB windows

1024th:
This improves HiDPI support in the classic UI XCB path.

Theme margins, overlays, action images, input window layout metrics, menu spacing, menu icons, separators, click regions, blur regions, and masks are now scaled with the current DPI.

This keeps the input window and XCB menu visually consistent on high-DPI displays instead of mixing scaled text with unscaled theme metrics.

Comparison on my KDE Wayland 175% scale setting, with Xft.dpi: 168 for Xwayland:
Wayland:
图片

Xwayland, before this fix:
图片

Xwayland, after this fix:
图片

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HiDPI support across classic menus and windows.
  * Corrected window sizing, rendering, shadows, blur effects, and popup placement on scaled displays.
  * Improved pointer hit testing and submenu interactions when display scaling is enabled.
  * Updated text and theme element scaling for clearer, more consistent appearance across different DPI settings.

### DIFF
--- a/src/ui/classic/xcbinputwindow.cpp
+++ b/src/ui/classic/xcbinputwindow.cpp
@@ -168,14 +168,10 @@ void XCBInputWindow::updatePosition(InputContext *inputContext) {
 }
 
 void XCBInputWindow::updateDPI(InputContext *inputContext) {
-    dpi_ = ui_->dpiByPosition(inputContext->cursorRect().left(),
-                              inputContext->cursorRect().top());
-
-    // Let Cairo device scale handle HiDPI so Pango keeps logical font sizes.
-    setFontDPI(-1);
+    auto dpi = ui_->dpiByPosition(inputContext->cursorRect().left(),
+                                  inputContext->cursorRect().top());
+    setScale(scaleForDPI(dpi));
 }
-
-double XCBInputWindow::scale() const { return scaleForDPI(dpi_); }
 
 void XCBInputWindow::update(InputContext *inputContext) {
     if (!wid_) {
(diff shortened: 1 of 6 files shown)


1024th:
Updated the PR to address the review comments:

- Made the Yoga prev/next button width match the painted rounded widths: https://github.com/fcitx/fcitx5/blob/bc071e271ee2dcc1904b7b35166c7377c74ec9c1/src/ui/classic/inputwindow.cpp#L1045-L1047
- Rounded natural-size tile dimensions instead of truncating scaled values: https://github.com/fcitx/fcitx5/blob/bc071e271ee2dcc1904b7b35166c7377c74ec9c1/src/ui/classic/theme.cpp#L334-L343
- Scaled shadow margins used for XCB input window positioning: https://github.com/fcitx/fcitx5/blob/bc071e271ee2dcc1904b7b35166c7377c74ec9c1/src/ui/classic/xcbinputwindow.cpp#L80-L115
- Rebuilt the blur region when the DPI scale changes, not only when the window size changes: https://github.com/fcitx/fcitx5/blob/bc071e271ee2dcc1904b7b35166c7377c74ec9c1/src/ui/classic/xcbinputwindow.cpp#L179-L237
- Used the bottom highlight margin for XCB menu item region height (this looks like an existing bug in the original code): https://github.com/fcitx/fcitx5/blob/bc071e271ee2dcc1904b7b35166c7377c74ec9c1/src/ui/classic/xcbmenu.cpp#L511-L517

wengxt:
The change could mostly be minimized to xcb part if you just simply downscale the X11 event coordinates and it shall be similar to wayland model.

1024th:
Thanks for the suggestion. I refactored the patch following this approach.

Theme layout and Pango metrics now stay in logical coordinates. The XCB backend uses Cairo device scale for rendering, scales only X11-specific window geometry and regions, and converts X11 pointer coordinates back to logical coordinates before passing them to the shared UI code.

1024th:
Changed the fixed 40-pixel offsets to `physicalFromLogical(40)`.

PR #1621, opened 2026-08-01
Add optional atomic surrounding-text replacement contract

dismonjames:
Adds an ABI-additive optional input-context interface for engines that need to replace validated surrounding text without splitting deletion and commit across client transactions. D-Bus exposes it only during ProcessKeyEventBatch; Wayland v1/v2 apply deletion and text in one protocol commit. Unsupported frontends do not advertise the interface.\n\nThis is being exercised by UniLume for zero-preedit Vietnamese input; additional focused protocol tests will follow before marking ready.


wengxt:
I doubt this solves any real problem.

1. most case, UI event is handled in singe thread
2. dbus message order is guaranteed, only tricky thing is whether you can ensure method call to work in the same order with signal

As long as event is handled in order, there's no need to do fancy things in the API.

PR #1624, opened 2026-08-03
Preserve current input method when updating input method group

wyu71:
* Keep the current input method when updating the input method group

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the existing default input method when replacing a group without a specified default.
  * Continued using the replacement group’s default when one is provided.
  * Added fallback behavior to select the first available input method when the configured default is unavailable.
  * Improved consistency when updating input method groups and retaining valid user preferences.

### DIFF
--- a/src/lib/fcitx/inputmethodmanager.cpp
+++ b/src/lib/fcitx/inputmethodmanager.cpp
@@ -352,7 +352,11 @@ void InputMethodManager::setGroup(InputMethodGroup newGroupInfo) {
                                    return !d->entries_.count(item.name());
                                });
     list.erase(iter, list.end());
-    newGroupInfo.setDefaultInputMethod(newGroupInfo.defaultInputMethod());
+    auto defaultInputMethod = newGroupInfo.defaultInputMethod();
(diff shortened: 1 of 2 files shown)


wyu71:
@wengxt Please review this.

wengxt:
This is wrong place to implement it, it should be done with input method manager where the group is changed.

wyu71:
@wengxt  Moved the preservation logic into `InputMethodManager::setGroup()` as suggested. Added regression coverage for both preserving an existing default and falling back when that input method is removed.

wyu71:
Added a separate invalid-default regression case. When the previous default `testim` is removed from the replacement group, `setGroup()` now falls back to the valid second entry `testim2`; the existing case still verifies preservation when `testim` remains.

wyu71:
The D-Bus setter was eagerly calling `setDefaultInputMethod("")`, which resolves the empty value to the second input method before `InputMethodManager::setGroup()` can preserve the existing default. Removed that normalization so the manager receives the unspecified default and performs the preservation or fallback consistently.

PR #1632, opened 2026-08-07
Preserve current input method when updating input method group

wyu71:
* Keep the current input method when updating the input method group

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved an input method group’s existing default when an update does not specify a replacement.
  * Retained valid replacement defaults and correctly selected a fallback input method when no default is provided.
  * Prevented group updates from unintentionally clearing the configured default input method.

* **Tests**
  * Added coverage for default preservation, replacement, validation, and fallback behavior.


wyu71:
Closing this duplicate PR because #1624 has already merged. The follow-up will contain only the D-Bus default-normalization removal.

PR #1633, opened 2026-08-07
Defer default input method selection to input method manager

wyu71:
* Follow-up to #1624

InputMethodManager::setGroup() already preserves an unspecified default or
falls back when it is invalid. The D-Bus path resolves an empty default to
the second input method before setGroup() runs, so remove that call.


wyu71:
@wengxt Please review this.

Issue #1584, opened 2026-06-18
Do not use a bundled plasma-wayland-protocol XML for window management


lbeltrame:
### Summary

The current one in the repository is 3 years old. And since it recently changed upstream (https://invent.kde.org/libraries/plasma-wayland-protocols/-/commit/bcb9d6ebcf7d43d2565295bec0f4e93d53ffc642) it can stop functioning (in fact, it did with the most recent Plasma  version). 

Fcitx5 should search for an installed version and use it as build dependency instead. Or at the very least, allow overriding the bundled version.

### Steps to Reproduce

Run fcitx5 against current plasma-wayland-protocols master. It will instantly quit the moment a window is changed.

### Expected Behavior

fctix5 should allow using a non-bundled plasma window management XML.

### Output of fcitx5-diagnose command

Not relevant to this issue.

Zamundaaa:
To clarify, it should not be used at all outside of plasmashell. We give no backwards compatibility guarantees for these private protocols.

CoelacanthusHex:
> To clarify, it should not be used at all outside of plasmashell. We give no backwards compatibility guarantees for these private protocols.

Unfornately, it may be the only way to implement this feature. There is 3 ways to be considered:
1. wlr-foreign-toplevel-management-unstable-v1, which is not implemented by KWin.
2. ext-foreign-toplevel-list-v1, which has no active state (There is a unmerged ext-foreign-toplevel-state to support this), and also not implemented by KWin.
3. kde-plasma-window-management, it's internal, but is the only possible way to do this.

I see https://invent.kde.org/plasma/kwin/-/work_items/311, hope we can have the standard way to do it soon. (Actually, I think this information may be better to be provided by compositor via zwp-input-method-v2 instead of the IME try to get this info via side way)

wengxt:
@Zamundaaa apparently fcitx's not the only one doing this: FYI

https://github.com/lxqt/lxqt-panel/blob/master/panel/backends/wayland/protocols/plasma-window-management.xml
https://github.com/dangvd/crystal-dock/blob/14e7d88589d8ab73f3d3984fe2ad0a2e96c07904/src/display/plasma_window_management.h#L202
https://github.com/Cairo-Dock/cairo-dock-core/blob/e8289c6eac32e4f411bf530fe09fa71e74c6ff06/src/implementations/proto/plasma-window-management.xml

Maybe you want to relax the version reset to a lower one around internet

lbeltrame:
[...]

PR #1634, opened 2026-08-07
Add cmake option to disable Plasma Wayland extensions

mojyack:
In Gentoo, starting from 5.1.20, the fcitx package introduced a dependency on plasma-wayland-protocols, which pulls in a lot of unwanted Qt dependencies.
While we could fix this on our side by turning USE_SYSTEM_PLASMA_WAYLAND_PROTOCOLS off, this extension is entirely unnecessary for non-Plasma users. Therefore, making it opt-out is a much cleaner solution.
This patch introduces a new CMake option, ENABLE_WAYLAND_PLASMA, for that purpose.

## Summary by CodeRabbit

* **New Features**
  * Added a build option to enable or disable Wayland Plasma integration.
  * Wayland Plasma support is enabled by default.

* **Improvements**
  * Plasma-specific monitoring and window-management components are included only when the feature is enabled.
  * Systems without Plasma support can now build with reduced integration requirements.


wengxt:
this is a pure xml protocol package which shouldn't pull in qt dependency, if it does, it's likely a packaging bug.

if don't want system use use the exisiting cmake option instead.

wengxt:
-DUSE_SYSTEM_PLASMA_WAYLAND_PROTOCOLS=Off

mojyack:
Thank you for replying.
Unfortunately plasma-wayland-protocols [actually pulls qt](https://github.com/KDE/plasma-wayland-protocols/blob/c421474708c26a409817c255e1c43939351444d8/CMakeLists.txt#L17).
At least, as long as install it proper way, qt dependencies are unavoidable.
I'll try fixing it on the Gentoo side.

wengxt:
@mojyack yeah, it seems need qt to run a ecm query qt. but I'd suggest you just use the cmake option above to avoid building the external one.

also fcitx5-configtool is qt anyway you might need qt eventually.

mojyack:
> also fcitx5-configtool is qt anyway you might need qt eventually.

I haven't used the configuration tool in a long time, I didn't realize it.

eli-schwartz:
> -DUSE_SYSTEM_PLASMA_WAYLAND_PROTOCOLS=Off

This does not skip building plasma integrations, it merely builds broken plasma integrations that per https://github.com/fcitx/fcitx5/issues/1584 will break everything and cause fcitx to abort.

You should remove fallback support for building broken plasma integrations that cause fcitx to abort. I've closed the Gentoo PR that proposes to build broken plasma integrations that cause fcitx to abort. Happy to look at a PR that makes fcitx not build any plasma integrations for non-plasma desktops, though.

wengxt:
actually.. I fixed that upstream though seems there's no new release yet.

but if you don't use qt thus no kwin, why would you bother anyway?

https://github.com/KDE/plasma-wayland-protocols/commit/39de7050bd60f104019193a494d819d5eae6ae9b

eli-schwartz:
I'm not really sure what you're trying to say.

"bundled-libs" / "vendored dependencies" are generally a bad idea and distros refrain from using them, for a variety of reasons including the fact that their tendency to get out of sync is liable to result in crashing software, as happened here. Publishing binaries that crash only for some users, is if anything much worse than binaries that consistently crash for all users.

If one doesn't use qt thus no kwin, one might wish to refrain from compiling code that is only for kwin. That's what this PR proposes to do. What is it that you are referring to when you say

> why would you bother anyway?

Bother with what?

wengxt:
wayland protocol is designed in a way to be backward compatible.

The original bug itself is KDE not following the standard/correct practice to update protocol file.

It is not even a problem about bundle or not. The implicit bundle will happen as soon as you build it.

The crash can be triggered for any client that is building with the old xml because of the change, including any application distirubuted on flatpak which may not rebuild or updated.

If the correct practice is followed, such crash won't even happen in the first place and that's the whole point of my fix.

With my fix to plasma-wayland-protocols, not using the system version will not crash too.

Bundle is actually very common in wayland world if you want to have a unreleased version with the support in binary.

for your use case to avoid qt, just use the cmake option and call it day. I don't want to have spaghetti ifdef when it's not necessary.

eli-schwartz:
> It is not even a problem about bundle or not. The implicit bundle will happen as soon as you build it.

One does not traditionally bundle a header-only library even if the resulting compiled binaries inline the results.

This is "just" a header-only library from that perspective.

> The crash can be triggered for any client that is building with the old xml because of the change, including any application distirubuted on flatpak which may not rebuild or updated.

My opinions about flatpak aren't flattering, so this comment of yours is insightful regarding why flatpak is bad.

> for your use case to avoid qt, just use the cmake option and call it day. I don't want to have spaghetti ifdef when it's not necessary.

"Spaghetti" is a pejorative you're using to make something very simple sound more complicated than it actually is. And on the flip side it makes things simpler for you when you don't need to be on the "vendor new updates" treadmill.

But personally I don't care, I'm fine with fcitx depending on qtbase -- it's a valid build dependency (protocols) for a non-optional feature (plasma extensions). I guess @mojyack cares though.

wengxt:
https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/601

wengxt:
Actually it's too common to bundle wayland protocols. I can gives tens of hundreds of examples.

plasma protocol is not the only one that fcitx bundles. there are also bunch of others include wlr ones, zwp_input_method_v2

What's funny is that even wlroots itself doesn't use their wlr-protocols package. 

https://gitlab.freedesktop.org/wlroots/wlroots/-/tree/master/protocol?ref_type=heads

The wayland eco is that the protocol generaeted code will be bundled in the binary the time as it's built.

It is the similar policy like linux kernel: never break user space.  As I said, it's a bug in kde upstream and it's fixed. And it's never in stable version.

PR #1658, opened 2026-09-06
fix *-import-addon.cpp overwrite on reconfigure

eagleoflqj:
Ref: https://cmake.org/cmake/help/git-stage/command/file.html
~`file(write` is not suitable for build input. `file(configure` needs cmake 3.18 which I think is OK since Ubuntu 24.04 is the [targeting system](https://groups.google.com/g/fcitx/c/1ZpjYVwVpqA) and it ships [cmake 3.28](https://launchpad.net/ubuntu/noble/+package/cmake).~

## Summary by CodeRabbit

* **Build and Packaging**
  * Improved generation of Fcitx5 addon import sources during builds.
  * Included the addon import template with the installed Fcitx5 utility package configuration files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.18)
 project(fcitx VERSION 5.1.22)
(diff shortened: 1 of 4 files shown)


wengxt:
It seems like you could just use configure_file and add a new template file.

PR #1664, opened 2026-09-08
classicui: expose candidate actions in X11 and Wayland

### LINKED ISSUE #1663: Expose actionable candidate operations from the Classic UI candidate window
## Summary
The Classic UI candidate window does not expose the candidate actions already provided by `ActionableCandidateList`. For example, the Pinyin engine can provide actions such as forgetting a candidate, pinning it as a custom phrase, and deleting a custom phrase, but these actions are not reachable from the desktop candidate window.

## Current behavior
- Left-click selects a candidate and the existing paging controls continue to work.
- Right-clicking a candidate is ignored by the Classic UI input window.
- The Wayland and X11 input-window handlers currently only forward candidate selection/paging events; they do not present `CandidateAction` entries.

## Expected behavior
When the selected candidate has actions, provide a context menu containing the actions returned by `ActionableCandidateList::candidateActions()`. The menu should:
- appear only for candidates for which actions are available;
- invoke `triggerAction()` with the original candidate and action id;
- preserve the existing left-click, wheel, keyboard, and paging behavior;
- support the action properties exposed by the API where the UI can represent them; and
- work consistently across the supported Classic UI backends, or clearly document any Wayland protocol limitation.

The implementation should reuse the existing `CandidateAction` API rather than hard-code Pinyin-specific operations. Related Wayland popup-surface discussion: #1654.

JohnXu22786:
## Summary

Expose the generic candidate actions provided by `ActionableCandidateList` from the Classic UI candidate window.

Both Classic UI backends now support right-clicking a candidate and choosing one of its actions:

- X11 uses the existing `Menu`/`XCBMenu` infrastructure.
- Wayland renders and handles the action menu inside the existing input popup surface, so it works with the current input-method popup protocol.

The UI invokes `triggerAction()` with the originating candidate and action id; no Pinyin-specific operation is hard-coded.

## Behavior

- A menu is shown only when the clicked candidate has at least one non-separator action.
- Separators, checkable state, and checked state are preserved where the backend can represent them.
- Temporary menu state is cleared when the input panel updates or the menu is dismissed.
- Existing left-click, wheel, touch, keyboard, and paging behavior remains unchanged when the menu is not active.
- The Wayland menu is sized and positioned within the popup surface, expanding the surface only when needed.

Closes #1663.

## Testing

- `cmake --build build-x11 -j2`
- `ctest --test-dir build-x11 --output-on-failure` (50/50 passed)
- `cmake --build build-wayland -j2`
- `ctest --test-dir build-wayland --output-on-failure` (50/50 passed)
- `clang-format --dry-run --Werror src/ui/classic/xcbinputwindow.cpp src/ui/classic/xcbinputwindow.h src/ui/classic/waylandinputwindow.cpp src/ui/classic/waylandinputwindow.h`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added right-click menus for candidate actions in the input panel.
  * Menus support toggles, separators, hover states, scrolling, and themed rendering.
  * On supported Wayland configurations, menus open in a separate popup surface.
  * Menu placement and sizing adapt to the input panel and display.
* **Bug Fixes**
  * Candidates with no actionable options no longer open an empty menu.
  * Existing left-click behavior for selecting candidates remains unchanged.
* **Usability**
  * Menus dismiss and refresh correctly when input or candidate content changes.

### DIFF
--- a/src/ui/classic/waylandinputwindow.cpp
+++ b/src/ui/classic/waylandinputwindow.cpp
@@ -42,6 +42,7 @@
 
 namespace fcitx::classicui {
 
+/** Initializes the Wayland input window and its event handlers. */
 WaylandInputWindow::WaylandInputWindow(WaylandUI *ui)
     : InputWindow(ui->parent()), ui_(ui), window_(ui->newWindow()) {
     menuContext_.reset(pango_font_map_create_context(fontMap_.get()));
@@ -133,6 +134,7 @@ WaylandInputWindow::WaylandInputWindow(WaylandUI *ui)
     initPanel();
 }
 
+/** Hides the candidate menu and discards its temporary state. */
 void WaylandInputWindow::clearCandidateMenu() {
     candidateMenuVisible_ = false;
     candidateMenuCandidate_ = nullptr;
@@ -147,6 +149,7 @@ void WaylandInputWindow::clearCandidateMenu() {
     candidateMenuHoveredIndex_ = -1;
 }
 
+/** Restores the input panel surface after a candidate menu closes. */
 void WaylandInputWindow::restorePanelSize() {
     if (!window_ || panelWidth_ <= 0 || panelHeight_ <= 0 ||
         (window_->width() == panelWidth_ &&
@@ -157,6 +160,7 @@ void WaylandInputWindow::restorePanelSize() {
     updateBlur();
 }
 
+/** Builds and shows the candidate action menu at the pointer position. */
 void WaylandInputWindow::showCandidateMenu(int x, int y) {
     auto dismiss = [this]() {
         if (!candidateMenuVisible_) {
@@ -310,6 +314,7 @@ void WaylandInputWindow::showCandidateMenu(int x, int y) {
     repaint();
 }
 
+/** Updates the candidate menu item under the pointer. */
 bool WaylandInputWindow::hoverCandidateMenu(int x, int y) {
     if (!candidateMenuVisible_) {
         return false;
@@ -330,6 +335,7 @@ bool WaylandInputWindow::hoverCandidateMenu(int x, int y) {
     return true;
 }
 
+/** Activates or dismisses the candidate menu after a left click. */
 void WaylandInputWindow::clickCandidateMenu(int x, int y) {
     if (!candidateMenuVisible_) {
         return;
@@ -363,6 +369,7 @@ void WaylandInputWindow::clickCandidateMenu(int x, int y) {
     repaint();
 }
 
+/** Paints the candidate action menu over the input panel. */
 void WaylandInputWindow::paintCandidateMenu(cairo_t *cr) {
     if (!candidateMenuVisible_) {
         return;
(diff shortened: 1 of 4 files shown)


wengxt:
I don't want to add it if it can only work on X11.

I think maybe it worth to thing if we could do something novel to support wayland

wengxt:
I mean.. isn't it looks too werid like this?

图片

eagleoflqj:
Here is how it looks on macOS

wengxt:
@eagleoflqj I know, but there's some technical issue on wayland to do it properly like a popup menu on wayland. Basically input panel surface can't (right now)  be attached with a xdg_popup. I'm thinking if we could avoid use popup but invent some "novel" way to display it.

JohnXu22786:
👌🏻I'll look further into this issue.

wengxt:
> 👌🏻I'll look further into this issue.

We can brainstorm some design first. The wayland thing is like dancing with shackles right now because of missing proper xdg popup support.

JohnXu22786:
I have updated this PR to address the Wayland UX concern.

For `wayland_v2`, candidate actions now use a separate `WaylandShmWindow` with its own `zwp_input_popup_surface_v2`. The candidate panel is no longer enlarged or painted over, and the menu state is cleared on input-panel updates and surface destruction. Existing hover, click, touch, leave, separator, checkable-action, and `triggerAction()` behavior is preserved.

The legacy `zwp_input_panel_v1` path intentionally keeps the existing panel behavior because that protocol does not provide an independent compositor-positioned popup role.

I rebased the fix onto the current PR head and pushed it to this branch. The complete `classicui` target was built with both X11 and Wayland enabled; `testmenu`, `testuserinterfacemanager`, and `testcandidatelist` pass, and clang-format plus `git diff --check` pass.

Please review the new Wayland popup implementation, especially compositor placement on `wayland_v2` and the compatibility fallback for legacy `wayland`.

JohnXu22786:
Follow-up: I also addressed the two remaining compatibility limitations from the previous update.

The menu is now an independent `wl_subsurface` of the input-panel surface instead of a second compositor-positioned `zwp_input_popup_surface_v2`. This gives the menu its own input surface while allowing client-side positioning relative to the clicked candidate. It is used for both `wayland_v2` and legacy `zwp_input_panel_v1`; the menu is placed below the candidate when there is room in the panel coordinate space, otherwise above it. The input panel surface is no longer enlarged and the menu is never painted into the candidate bar.

This also avoids relying on compositor-chosen popup placement for the action menu. The change is in commit `e91089a4`, pushed to this PR branch. I rebuilt `classicui` with X11 and Wayland enabled; `testmenu`, `testuserinterfacemanager`, and `testcandidatelist` pass, and clang-format plus `git diff --check` pass.

Please review the subsurface approach on both the v2 and legacy v1 protocols, especially compositor input handling for the child surface.

wengxt:
I don't feel wl sub surface would solve any problem, because I believe sub surface is not allow to show window out side parent. Can you provide screen shot?

PR #1679, opened 2026-09-18
add OrderedAnnotation for List

eagleoflqj:
Make ordered/unordered list config distinguishable

## Summary by CodeRabbit

* **New Features**
  * Configuration descriptions can now indicate when list values preserve their order.
  * Multiple configuration annotations, such as ordering and tooltips, can be combined and displayed together.
  * Ordering metadata is now supported for long-press candidates, input method switching keys, and spell provider preferences.

* **Tests**
  * Added coverage confirming ordered and tooltip metadata appears correctly in configuration descriptions.


wengxt:
TBH, the ordering has always been available, not sure what does this want to achieve here.

For example, punctuation config's order also matters.

I think it's more applied to key list? so maybe just be more specific that this applies to key list, since normally key list is unordered so in configtool there's not even button to re-order key list. Other list should be mostly fine..

eagleoflqj:
> not sure what does this want to achieve here

1. From UI it confuses people whether order counts for KeyList. It's not intuitive that 以词定字 and im selector's first key means first character/im.
2. Sort buttons (↑) are unnecessary for unordered key list. Reducing UI elements is helpful, considering there are already too many (and will be more if we keep adding configuration) elements in a single page.

eagleoflqj:
> so in configtool there's not even button to re-order key list

Right, and this is something that needs improvement. The 三选 and 以词定字 have the same UI, with completely different semantics.



> Other list should be mostly fine

Agree, let me see if there's a way to restrict to key list ...

eagleoflqj:
Well actually, there are other types that order doesn't count for list. For example I have a per-program input method config, which is unordered, and a per-program vim-mode config

wengxt:
My point is that you can make it more specific, only applies to KeyList.

